PR #20, opened 2015-09-19
Kotlin M13 support.

### LINKED ISSUE #19: Kotlin m13

yoavst:
I've converted the project to Kotlin M13. (fixes #19 )
In some places it uses `lazy(LazyThreadSafetyMode.NONE)` instead of just `lazy`. That is because the automatic fix on all the project. I think it works the same as `lazy`, so you can replace it later.
Also, in order to remove a warning, I've added `Method.PATCH` to the when expression. However, `Method.PATCH` is inaccessible as a method, so it doesn't actually change something.

```kotlin
private fun setDoOutput(connection: HttpURLConnection, method: Method) {
         when (method) {
            Method.GET, Method.DELETE -> connection.doOutput = false
            Method.POST, Method.PUT, Method.PATCH -> connection.doOutput = true
         }
     }
```

Also I've run the tests, all pass. So it looks like no functionality were changed.

### DIFF
--- a/fuel/src/main/kotlin/fuel/core/Request.kt
+++ b/fuel/src/main/kotlin/fuel/core/Request.kt
@@ -45,7 +45,7 @@ public class Request {
     }
 
     //underlying task request
-    val taskRequest: TaskRequest by Delegates.lazy {
+    val taskRequest: TaskRequest by lazy(LazyThreadSafetyMode.NONE) {
         when (type) {
             Type.DOWNLOAD -> DownloadTaskRequest(this)
             Type.UPLOAD -> UploadTaskRequest(this)
@@ -189,9 +189,9 @@ public class Request {
     public fun responseJson(handler: Handler<JSONObject>): Unit = response(Request.jsonDeserializer(), handler)
 
     //object
-    public fun <T> responseObject(deserializer: ResponseDeserializable<T>, handler: (Request, Response, Either<FuelError, T>) -> Unit): Unit = response(deserializer, handler)
+    public fun <T: Any> responseObject(deserializer: ResponseDeserializable<T>, handler: (Request, Response, Either<FuelError, T>) -> Unit): Unit = response(deserializer, handler)
 
-    public fun <T> responseObject(deserializer: ResponseDeserializable<T>, handler: Handler<T>): Unit = response(deserializer, handler)
+    public fun <T: Any> responseObject(deserializer: ResponseDeserializable<T>, handler: Handler<T>): Unit = response(deserializer, handler)
 
     public fun cUrlString(): String {
         val elements = arrayListOf("$ curl -i")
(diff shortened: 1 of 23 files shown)


kittinunf:
@yoav. That's awesome!!! I was skimming through changes and everything is good! :+1. Let me build on my machine once again and I ll merge this in asap. 



—

fuel/src/main/kotlin/fuel/core/Request.kt
```diff
@@ -189,9 +189,9 @@ public class Request {
     public fun responseJson(handler: Handler<JSONObject>): Unit = response(Request.jsonDeserializer(), handler)
 
     //object
-    public fun <T> responseObject(deserializer: ResponseDeserializable<T>, handler: (Request, Response, Either<FuelError, T>) -> Unit): Unit = response(deserializer, handler)
+    public fun <T: Any> responseObject(deserializer: ResponseDeserializable<T>, handler: (Request, Response, Either<FuelError, T>) -> Unit): Unit = response(deserializer, handler)
```
kittinunf: Do you have a specific reason why we need to give `T` an upper bound with `Any`?

kittinunf:
Document for `lazy` says 

``` Kotlin
/**
 * Creates a new instance of the [Lazy] that uses the specified initialization function [initializer]
 * and the default thread-safety mode [LazyThreadSafetyMode.SYNCHRONIZED].
**/
```

So it seems like plain `lazy` would automatically be synchronized. Meanwhile, the `lazy(LazyThreadSafetyMode.NONE)` wouldn't.  

Perhaps, just to be safe, I guess we should stick with the non-synchronized version as I don't think it is needed.

yoavst:
I didn't know about the lazy. 

About generics.

Maybe @mplatvoet would be able to explain it better since he is the developer of [Kovenant](https://github.com/mplatvoet/kovenant).

IDEA's Kotlin M13 plugin shows a warning if `T` doesn't extend `any` when you use `T?`.
 If the methods I changed can return `T?`, then you better keep it `T : Any` but change the return type to `T?`

kittinunf:
OK. I think you are right.

BTW, I think I will create a proper branch for this because I would like to fix #21 created by him. Would you mind create another PR against my 0.53 branch? Then I would fix #21 together with it.

yoavst:
I'm not a git pro, so I'll have to reapply all the changes. I can do another commit with the sealed change, lazy fix, and nullability fix for deserializer if needed.

kittinunf:
I created this for you to merge your M13 fix in. #22

kittinunf:
C'mon .. I would love to not let you do this alone. Lemme help :) I will fix other small things as well.

mplatvoet:
I think the lazy instance *should* be thread safe in most cases here. Same goes for `readWriteLazy` which is used for instance in headers. The resulting Request instance can be shared amongst threads and I think people will. Only if you can guarantee that a lazy can only be initialised and read by one thread you should use the unsynced variant.

yoavst:
Also, what about nullability in generics for `Deserializable`?
Should it be able to return null?

PR #22, opened 2015-09-19
Support Kotlin M13

kittinunf:
- [x] Mark Either as sealed.
- [x] Update sample to reflect Kotlin M13.
- [x] M13 Support by @yoavst, :+1: 
- [x] Update Readme.

fix #21, fix #19, close #20

PR #23, opened 2015-09-19
0.53

### LINKED ISSUE #21: Make Either sealed
M13 introduces the concept of [sealed classes](http://kotlinlang.org/docs/reference/classes.html#sealed-classes) which improves pattern matching so you are not required to use an `else` case anymore.

yoavst:
A complement for 0.53. Fix #21, #19 and complete #22


kittinunf:
Looks great!! I will get this merged in. Thanks! @yoavst

PR #26, opened 2015-10-02
M14 support

yoavst:
I didn't mark `Fuel.get` as operator, since it isn't. It will still show warning, but if I understand right the warning will be not shown on M15, and users will not be able to use it as operator.


kittinunf:
It looks great! I have my own branch locally but I haven't pushed it up. I am happy to accept this.

PR #33, opened 2015-11-02
Non android support

### LINKED ISSUE #18: Non Android support
Is it planned? Currently it requires to use robolectric to test api calls.
From a quick view (running without roboelectric) you use android's Uri class. Is there any alternative?

yoavst:
Fix #18 
The changes are:
* Remove callback to UI by default. A `fuel-android` module will be needed. It will have to set [AndroidMainThreadExecutor](https://github.com/kittinunf/Fuel/blob/master/fuel/src/main/kotlin/fuel/util/AndroidMainThreadExecutor.kt) as default.
```kotlin
Manager.callbackExecutor = AndroidMainThreadExecutor()
```
* Remove `JSONObject` support. also for `fuel-android`.
* `Request.response` is no longer internal so it will support extension (currently used for testing json). 
* authenticating is using reflection to support it on both jvm and android. I've never done authenticating so you better check it still works. My old solution was copy paste android class to project, but I couldn't compile it. 
* Switch from `Uri` to `URI`, now some address are invalid, but I guess they would have been invalid anyway.


fuel/src/main/kotlin/fuel/core/Deserializable.kt
```diff
@@ -36,7 +36,7 @@ public interface ResponseDeserializable<out T : Any> : Deserializable<T> {
 
 }
 
-internal fun <T: Any, U : Deserializable<T>> Request.response(deserializable: U, handler: (Request, Response, Either<FuelError, T>) -> Unit) {
+/*internal*/ fun <T: Any, U : Deserializable<T>> Request.response(deserializable: U, handler: (Request, Response, Either<FuelError, T>) -> Unit) {
```
kittinunf: Remove this please :)

fuel/src/main/kotlin/fuel/core/Deserializable.kt
```diff
@@ -44,7 +44,7 @@ internal fun <T: Any, U : Deserializable<T>> Request.response(deserializable: U,
     })
 }
 
-internal fun <T: Any, U : Deserializable<T>> Request.response(deserializable: U, handler: Handler<T>) {
+/*internal*/ fun <T: Any, U : Deserializable<T>> Request.response(deserializable: U, handler: Handler<T>) {
```
kittinunf: Remove this also

fuel/src/main/kotlin/fuel/core/Deserializable.kt
```diff
@@ -52,7 +52,7 @@ internal fun <T: Any, U : Deserializable<T>> Request.response(deserializable: U,
     })
 }
 
-internal fun <T: Any, U : Deserializable<T>> Request.response(deserializable: U,
+/*internal*/ fun <T: Any, U : Deserializable<T>> Request.response(deserializable: U,
```
kittinunf: This one should be `private`

kittinunf:
The rest is fine, thank you so much @yoavst :cherry_blossom: 

We probably need `fuel-android` module before we have another release, I think this one deserves a 0.6 :)

yoavst:
Done. I'll work on `fuel-android` tomorrow. How do you think the integration should be?
```Kotlin
// 1. Extension method
Fuel.configAndroid()
// 2. Object method
FuelAndroid.config()
// 3. Just the handler
Manager.callbackExecutor = AndroidMainThreadExecutor()
```
It should also include `JSONObject` support.

kittinunf:
I am leaning toward the least hassle as possible? How about we just configure the `callbackExecutor` for client and see how it goes there?

yoavst:
In order to that code has to be executed, so you can't auto config automatically. A method must be called.

PR #38, opened 2015-12-01
0.66

kittinunf:
- [x] Move to `Result`. Clear depart from `Either`
- [x] Cancel
- [x] Update README.md


kittinunf:
We also need to work on Readme.md.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Deserializable.kt
```diff
@@ -58,21 +61,18 @@ private fun <T: Any, U : Deserializable<T>> Request.response(deserializable: U,
     taskRequest.apply {
         successCallback = { response ->
 
-            val deliverable: Either<Exception, T> =
+            val deliverable: Result<T, Exception> =
                     try {
-                        Either.Right(deserializable.deserialize(response))
```
kittinunf: This supposed to use `Result` 's semantic which could get rid of try catch.

PR #41, opened 2015-12-24
Synchronous call support

bmsantos:
Introduced synchronous call support by adding sync() method to Request class.
When called during the flow API, a flag will be set to indicate that sync mode is to be use.
This flag is checked after the request submission and, if flagged, then future will make use of [get](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html#get()) method to wait for request completion.

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Deserializable.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Deserializable.kt
@@ -82,9 +82,7 @@ private fun <T : Any, U : Deserializable<T>> Request.response(deserializable: U,
 
     submit(taskRequest)
 
-    if (this.syncMode) {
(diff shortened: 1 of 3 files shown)


yoavst:
Looks good. @kittinuf?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Deserializable.kt
```diff
@@ -81,5 +82,9 @@ private fun <T : Any, U : Deserializable<T>> Request.response(deserializable: U,
 
     submit(taskRequest)
 
+    if (this.syncMode) {
```
kittinunf: Can you omit `this` so it is consistent with everywhere else?

bmsantos:
Not a problem. I'm away now but I'll change it as soon as i can.

kittinunf:
Looks great, thanks!

PR #43, opened 2016-01-18
Support for blocking thread mode

yoavst:
## Background 
`Fuel` was built on the concept of an Async HTTP networking library. Later, the `sync()` method was added to support blocking the current thread.

* Fuel supports synchronous call by calling `sync` before `response`. 
``` Kotlin
var data: String? = null
//the call will block until the http call finished
Fuel.get("http://httpbin.org/get").sync().responseString { req, res, result ->
    val (d, e) = result
    data = d
}

//do something with data
```

However, it's still complicated to use with an asynchronous library like [Koventant](http://kovenant.komponents.nl/):
```kotlin
public fun Request.promise(): Promise<Pair<Response, ByteArray>, Exception> {
    val deferred = deferred<Pair<Response, ByteArray>, Exception>()
    response { request, response, result ->
        when (result) {
            is Result.Failure -> deferred.reject(result.error!!)
            is Result.Success -> deferred.resolve(Pair(response, result.value!!))
        }
    }
    return deferred.promise
}
```
This method uses `Fuel`'s async support rather than `Kovenant`'s one, which is bad for the developer that configs `Kovenant`.

## Proposal

`Fuel`'s `Request` has now 3 new methods:
```kotlin
request.response() -> Triple<Request, Response, ByteArray>
request.responseString() -> Triple<Request, Response, String>
request.responseObject(deserializer: ResponseDeserializable<T) -> Triple<Request, Response, T>
```

Since `Triple` is a data class, it can be unpacked:
```kotlin
val (request, response, data) = "http://www.google.com".httpGet().responseString()
```

Those methods block the current thread so they can be easily wrapped with `Kovenant`:
```kotlin
task { "http://www.google.com".httpGet().responseString() }
```

## Implementation 
`TaskRequest` becomes `Callable<Response`. For async request it is being wrapped by `AsyncTaskRequest` which is responsible for notifying the callbacks.


kittinunf:
I don't see any concerns. Can you also update README to reflect our new blocking mode? Once it is done, we shall merge this in.

kittinunf:
Looking great. You can merge anytime you wish.

yoavst:
Merged.

Issue #52, opened 2016-02-11
TimeoutExcpetion is synchronized requests


yoavst:
### `Sync()` method and timeout
Given the following code:
```kotlin
Fuel.get("https://www.google.co.il").sync().timeout(5000).responseString { request, response, result ->
            println("Received; Start sleeping for 5 seconds.")
            Thread.sleep(5000)
            println("Done")
        }
```

The result is:
```
Received; Start sleeping for 5 seconds.
Exception in thread "main" java.util.concurrent.TimeoutException
    at java.util.concurrent.FutureTask.get(FutureTask.java:205)
    at com.github.kittinunf.fuel.core.DeserializableKt.response(Deserializable.kt:78)
    at com.github.kittinunf.fuel.core.DeserializableKt.response(Deserializable.kt:35)
    at com.github.kittinunf.fuel.core.Request.responseString(Request.kt:223)
    at com.github.kittinunf.fuel.core.Request.responseString$default(Request.kt:222)
    at com.yoavst.testing.fuel.SyncMethodBug.run(SyncMethodBug.kt:8)
    at com.yoavst.testing.MainKt.main(main.kt:7)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:497)
    at com.intellij.rt.execution.application.AppMain.main(AppMain.java:144)
Done
```

### Bug info
This [Code](https://github.com/kittinunf/Fuel/blob/master/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Deserializable.kt#L78) is responsible on this bug. The timeout is being applied twice for `sync()`, once in [`HttpClient`](https://github.com/kittinunf/Fuel/blob/master/fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt#L20) and here.

So the question is - is it duplicated for a reason, or it is just a bug?

yoavst:
closing this as there is no `sync()` method anymore.

PR #53, opened 2016-02-13
Test mode support, fix for #52

yoavst:
@kittinunf What we've talked about on Slack.


yoavst:
@kittinunf ?

kittinunf:
OK good to go!

PR #58, opened 2016-03-07
Update 'Synchronous Call' documentation in README.

glureau:
Just lost 5 minutes searching the sync() method.
Do not hesitate to cancel this PR and improve the documentation on your side if you want to (it's just an easy-fix for you anyway).

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,16 +355,10 @@ request.cancel()
 ```
 
 ### Synchronous Call 
-* Fuel supports synchronous call by calling `sync` before `response`. 
+* Fuel supports synchronous call by calling the `response()` methods without handler. 
 ``` Kotlin
-var data: String? = null
 //the call will block until the http call finished
-Fuel.get("http://httpbin.org/get").sync().responseString { req, res, result ->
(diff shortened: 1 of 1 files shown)


kittinunf:
Actually, we have 2 ways to force Fuel to be asynchronous.

First, use `sync()`, second use the `response()` method. Both are async

The difference is the former preserves interface to be a callback-ish kinda way but in synchronous mode, such as 

``` Kotlin
Fuel .... .sync().response { request, response, result ->
}
```

Meanwhile, the latter yields interface such as, 

``` Kotlin
val (request, response, data) = Fuel ... .response()
```

I am all in for adding/improving README, but you might wanna change things a bit to be perfectly accurate.

yoavst:
Actually, I removed the `sync()` method for `response()`. 

For testing you can still use `Fuel.testMode {} `

glureau:
Didn't see the "Blocking" section before!

glureau:
Looks like https://github.com/kittinunf/Fuel/pull/61 already includes the modification.
Closing this PR.

PR #60, opened 2016-03-08
Make sync version support Result<Data, FuelError>

yoavst:
Now it has the same API is async version, allows you query the `Request` object even if the request has failed.


kittinunf:
I prefer to have `Result` wrapping the Result at the end of the pipeline. Can you refactor a bit more so that it won't spread inside Fuel core?

yoavst:
So in which other way do you want the task to return a `Response` or `FuelError`? 
`Pair<Reaponse?, FuelError?>`?

kittinunf:
Is there any case that simple `Response` doesn't cover?

PR #61, opened 2016-03-08
Make sync version support Result<Data, FuelError>

yoavst:
Now it has the same API is async version, allows you query the Request object even if the request has failed.

Note: #60 was deleted since I removed that commit of it.


kittinunf:
This looks exactly what I wish it could be implemented. GTM :+1:

PR #63, opened 2016-03-11
Fix Encoding issue

yoavst:
Should fix some encoding issues, check the added tests for extra details.

Example for URLs that works now:
`https://www.example.com/files/%D7%98%D7%A7%D7%A1%D7%98%20%D7%91%D7%A2%D7%91%D7%A8%D7%99%D7%AA%201%203.txt`

`https://www.example.com/files/טקסט בעברית 1 3.txt`


kittinunf:
I am not sure why this work but if it solves the problem so be it.

yoavst:
The 'toAsciiString()` fix the problem with non ascii string. 
About the second fix, I have no idea 😐

kittinunf:
:shipit:

PR #70, opened 2016-03-30
Feature/json body

pawin:


yoavst:
but what if we want to use another content type?

pawin:
I'm finding how to create enum with custom value, is it possible to do that?

kittinunf:
I think It is impossible to cover all the case as `Content-Type` can be just about anything.

http://greenbytes.de/tech/webdav/draft-ietf-httpbis-p2-semantics-26.html#header.content-type

After careful consideration, we have to tackle this in [Encoding](https://github.com/kittinunf/Fuel/blob/master/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt) class instead. We need a enum to specify how we encode the parameters.

yoavst:
It can be implemented as Interceptor.

PR #74, opened 2016-04-16
Add responseJson in sync mode in Android module, fix #71

### LINKED ISSUE #71: responseJson() can not be used in block mode
```
val(request, response, reulst) = "http://test.com".httpGet().responseJson()
```

kittinunf:
Just a single line to support `responseJson()`.


kittinunf:
I think this should be straight forward. I will merge this in.

PR #76, opened 2016-04-19
Rename to FuelManager

### LINKED ISSUE #59: Manager
Hi guys!
First of all, well done for this very convenient library! Really cool to use :)
But I'm not super fan of the Manager, I don't like the static but it's super convenient here so ok, but the name...
Is it possible to rename it FuelManager or something like this?

yoavst:
Fix #59 again


kittinunf:
Looks good! You can merge as soon as you go.

PR #77, opened 2016-04-19
Fix headers

yoavst:
Beside making few `var`s to `val`s, it fixes the `content-type` problem.

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
@@ -19,7 +19,7 @@ class Encoding : Fuel.RequestConvertible {
     var encoder: (Method, String, List<Pair<String, Any?>>?) -> Request = { method, path, parameters ->
         var modifiedPath = path
         var data: ByteArray? = null
-        var headerPairs: MutableMap<String, Any> = hashMapOf("Accept-Encoding" to "compress;q=0.5, gzip;q=1.0")
+        val headerPairs: MutableMap<String, Any> = defaultHeaders()
         if (encodeParameterInUrl(method)) {
             var querySign = ""
             val queryParamString = queryFromParameters(parameters)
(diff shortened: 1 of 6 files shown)


fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
```diff
@@ -74,9 +75,13 @@ class Encoding : Fuel.RequestConvertible {
     private fun queryFromParameters(params: List<Pair<String, Any?>>?): String {
         return params?.let {
             params.filterNot { it.second == null }
-                    .mapTo(arrayListOf<String>()) { "${it.first}=${it.second}" }
+                    .mapTo(ArrayList()) { "${it.first}=${it.second}" }
```
kittinunf: We should stick with Kotlin types. Using `mutableListOf()` should be better option.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
```diff
@@ -34,7 +34,7 @@ class Request : Fuel.RequestConvertible {
     lateinit var url: URL
     var httpBody: ByteArray = ByteArray(0)
 
-    var httpHeaders = hashMapOf<String, String>()
+    val httpHeaders: MutableMap<String, String> = hashMapOf<String, String>()
```
kittinunf: Explicit type declaration is not needed.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/AsyncTaskRequest.kt
```diff
@@ -9,21 +9,22 @@ class AsyncTaskRequest(val task: TaskRequest) : TaskRequest(task.request) {
     var failureCallback: ((FuelError, Response) -> Unit)? = null
 
     override fun call(): Response {
-        try {
+        return try {
```
kittinunf: Nice 👍

kittinunf:
If you using rebase against `master`, this commit (737e7c5) is not needed.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
yoavst: Yes, it could be replaced with `mutableMapOf`, I want it to be kotlin type.

kittinunf: Yeah, that would be ideal. Can you help me fix this as well?

kittinunf:
It all looks good, after you fix my concerns.

kittinunf:
OK .. you are GTG.

PR #82, opened 2016-04-27
First commit for ReadTimeout option

carmas123:
First preview branch for additional read timeout param.


kittinunf:
Can you also add test case for that as well?

PR #86, opened 2016-05-10
`PathStringConvertible`as part of the public API only

yoavst:
Currently we have methods that do the same for `String` and for `PathStringConvertible`. This pull request removes the duplicated methods, making `PathStringConvertible` only a part of the public API, and not a type which is used internally.

### DIFF
--- a/fuel/src/test/kotlin/com/github/kittinunf/fuel/BlockingRequestTest.kt
+++ b/fuel/src/test/kotlin/com/github/kittinunf/fuel/BlockingRequestTest.kt
@@ -152,7 +152,7 @@ class BlockingRequestTest : BaseTestCase() {
 
     @Test
     fun httpGetRequestWithPathStringConvertible() {
-        val (request, response, data) = manager.request(Method.GET, HttpsBin.USER_AGENT.path).responseString()
+        val (request, response, data) = manager.request(Method.GET, HttpsBin.USER_AGENT).responseString()
 
         assertThat(request, notNullValue())
         assertThat(response, notNullValue())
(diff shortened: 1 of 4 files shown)


fuel/src/test/kotlin/com/github/kittinunf/fuel/BlockingRequestTest.kt
```diff
@@ -152,7 +152,7 @@ class BlockingRequestTest : BaseTestCase() {
 
     @Test
     fun httpGetRequestWithPathStringConvertible() {
-        val (request, response, data) = manager.request(Method.GET, HttpsBin.USER_AGENT).responseString()
+        val (request, response, data) = manager.request(Method.GET, HttpsBin.USER_AGENT.path).responseString()
```
kittinunf: Should this still work without the change? I should correctly resolve to use the `PathStringConvertible` version.
`

fuel/src/test/kotlin/com/github/kittinunf/fuel/BlockingRequestTest.kt
```diff
@@ -200,7 +200,7 @@ class BlockingRequestTest : BaseTestCase() {
         val headerValue = "application/json"
         manager.baseHeaders = mapOf(headerKey to headerValue)
 
-        val (request, response, data) = manager.request(Method.POST, HttpsBin.POST, listOf("email" to "foo@bar.com")).responseString()
+        val (request, response, data) = manager.request(Method.POST, HttpsBin.POST.path, listOf("email" to "foo@bar.com")).responseString()
```
kittinunf: Here too.

fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestTest.kt
```diff
@@ -256,7 +256,7 @@ class RequestTest : BaseTestCase() {
         var data: Any? = null
         var error: FuelError? = null
 
-        manager.request(Method.GET, HttpsBin.USER_AGENT).responseString { req, res, result ->
+        manager.request(Method.GET, HttpsBin.USER_AGENT.path).responseString { req, res, result ->
```
kittinunf: Here too?

kittinunf:
Looks good for me :shipit:

PR #87, opened 2016-05-11
Fix Read Timeout param

carmas123:


kittinunf:
@yoav do you have any concerns?

yoavst:
Not, looks good.

PR #110, opened 2016-08-23
Redirection following can now be disabled for relative urls. With test.

spand:
HTTPUrlConnection is following redirects by default. It seems the idea is that this is to be handled by Fuel itself.

I couldnt run the other tests as typicode.com returns 403 for me, but the new test case failed before and doesnt now.

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/RedirectionInterceptor.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/RedirectionInterceptor.kt
@@ -5,6 +5,8 @@ import com.github.kittinunf.fuel.core.Encoding
 import com.github.kittinunf.fuel.core.FuelError
 import com.github.kittinunf.fuel.core.Request
 import com.github.kittinunf.fuel.core.Response
+import java.net.MalformedURLException
+import java.net.URL
 import javax.net.ssl.HttpsURLConnection
 
 class RedirectException() : Exception("Redirection fail, not found URL to redirect to")
@@ -18,7 +20,13 @@ fun redirectResponseInterceptor() =
                     if (redirectedUrl != null && !redirectedUrl.isEmpty()) {
                         val encoding = Encoding().apply {
                             httpMethod = request.httpMethod
-                            urlString = redirectedUrl[0]
+                            urlString = try {
+                                URL(redirectedUrl[0]).toString()
+                            } catch (e: MalformedURLException){
+                                // Maybe its a relative url
+                                val serverPart = request.url.protocol + "://" + request.url.authority
+                                URL(serverPart + redirectedUrl[0]).toString()
+                            }
                         }
                         Fuel.request(encoding).response().second
                     } else {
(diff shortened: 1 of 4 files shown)


kittinunf:
Do you think whether it is more future-proof by setting the true, and using the facility provided by `HttpURLConnection` instead? It seems that using stuff provided by framework is better.

spand:
I think the best action would be to disable redirection on `HttpURLConnection` as my PR does and keep using the framework.. So I think we are in agreement :-)

spand:
So redirectResponseInterceptor actually didnt work on relative redirects.. ie `Location: /news/1`. It does now and passes the test also.

spand:
@kittinunf The commits in this PR make Fuel work for my purpose now. Let me know if you need anything to get this merged.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/RedirectionInterceptor.kt
```diff
@@ -1,15 +1,13 @@
...
-import com.github.kittinunf.fuel.core.Response
+import com.github.kittinunf.fuel.core.*
+import java.net.MalformedURLException
+import java.net.URL
 import javax.net.ssl.HttpsURLConnection
 
 class RedirectException() : Exception("Redirection fail, not found URL to redirect to")
 
-fun redirectResponseInterceptor() =
+fun redirectResponseInterceptor(manager: FuelManager) =
```
kittinunf: Seems like what we need is a baseUrl in `request.url`. Could we pass just that along and still using the `Fuel.request` version. Passing the whole manager in is a bit overkill.

kittinunf:
Thank you for your lovely PR. I have a single comment for you and the rest looks fine. 

BTW, I see tests fail because we can no longer call typicode.com. I will address the test issue in another PR.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/RedirectionInterceptor.kt
spand: The reason I pass the whole manager is that the subsequent requests will then be using the same baseHeaders, keystore, socketFactory, hostnameVerifier, executor etc. It seems to me that this is what one would expect. Otherwise any redirected requests will all be handled by FuelManager.instance.

kittinunf: Ah I see. The only thing I am kind of worried is reference cycle. It is like the `manager` is holding the `interceptor` and at the same time `interceptor` is holding reference back to `manager` in the lambda. At this point, it seems vague to me when things get destroyed.

spand: It looks fine to me. 
- If the `interceptor` is removed from the `manager` then the `interceptor` will become garbage even if it holds a reference to the `manager`
- If the `manager` is no longer referenced both will become garbage even if there is a reference cycle.

The java garbage collector has no problem with cycles, unlike what I believe is the case with iOS or Python. If this is your background maybe this stackoverflow question will make it more clear: http://stackoverflow.com/questions/1910194/how-does-java-garbage-collection-work-with-circular-references

kittinunf: Thanks!

kittinunf:
I will merge this in. Thank you so much @spand.

spand:
Thank you !

Issue #113, opened 2016-09-10
Boundary should keep same in one upload request


huiyiqun:
There are two boundary in one upload request:

[1] https://github.com/kittinunf/Fuel/blob/22734da6995dfa767ba4f73bcb36381016ccd83a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt#L34
[2] https://github.com/kittinunf/Fuel/blob/484d03bd6ae4740e659ab2a69ceebb88be94ec5e/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/UploadTaskRequest.kt#L18

They all take the current system time and convert it to hex string. However, as `currentTimeMills` is called twice, there is a high possibility that the return values are different.

However, according to [RFC7578](https://tools.ietf.org/html/rfc7578), they should keep same so that the server could recognize the form.

huiyiqun:
[Here](https://github.com/huiyiqun/Fuel/commit/ad9e15f35a951673a01cd8e6f2a98c4d50dce749) is a simple fixup.

kittinunf:
Can you please address my concern in the PR? Then we can merge this! :)

kittinunf:
Fixed!

PR #114, opened 2016-09-10
fix upload task

huiyiqun:
The current code has following problem:
1. the boundary in the content-type and that in body may be different. See #113 .
2. the CRLF is escaped that CRLF in packet is \x5c \x72 \x5c \x6e instead of \x0d \x0a .

This PR fixes above problems and makes upload works correctly.

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/UploadTaskRequest.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/UploadTaskRequest.kt
@@ -14,7 +14,7 @@ class UploadTaskRequest(request: Request) : TaskRequest(request) {
 
     val BUFFER_SIZE = 1024
 
-    val CRLF = "\\r\\n"
+    val CRLF = "\r\n"
     /*val boundary = System.currentTimeMillis().toHexString()*/
     val boundary = request.httpHeaders["Content-Type"]!!.split("=", limit=2)[1]
 
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/UploadTaskRequest.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/UploadTaskRequest.kt
@@ -15,7 +15,6 @@ class UploadTaskRequest(request: Request) : TaskRequest(request) {
     val BUFFER_SIZE = 1024
 
     val CRLF = "\r\n"
-    /*val boundary = System.currentTimeMillis().toHexString()*/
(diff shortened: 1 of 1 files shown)


fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/UploadTaskRequest.kt
```diff
@@ -14,8 +14,8 @@ class UploadTaskRequest(request: Request) : TaskRequest(request) {
 
     val BUFFER_SIZE = 1024
 
-    val CRLF = "\\r\\n"
-    val boundary = System.currentTimeMillis().toHexString()
+    val CRLF = "\r\n"
+    val boundary = request.httpHeaders["Content-Type"]!!.split("=", limit=2)[1]
```
kittinunf: I am a bit hesitant about using `!!`. It would crash if `httpHeaders` doesn't contain `"Content-Type"`. Can you do sth about that?

huiyiqun: Sorry for long time not here.

I'm not sure what's the correct behavior when there are no "Content-Type".

According to the RFC of `multipart/form-data`, the parameter `boundary` in "Content-Type" is required to parse the body. If there is no "Content-Type", the body makes completely no sense.

huiyiqun: The `!!` here is intended to mean that "Content-Type" must be here or something terrible has happed.

I guess a more verbose error is prefered, but I do not know kotlin very well and am not sure how to do that.

kittinunf: In the spec, it says it **SHOULD** be there

> Content-Type specifies the media type of the underlying data.
Content-Encoding may be used to indicate any additional content
codings applied to the data, usually for the purpose of data
compression, that are a property of the requested resource. There is
no default encoding.

> Any HTTP/1.1 message containing an entity-body SHOULD include a
Content-Type header field defining the media type of that body. If
and only if the media type is not given by a Content-Type field, the
recipient MAY attempt to guess the media type via inspection of its
content and/or the name extension(s) of the URI used to identify the
resource. If the media type remains unknown, the recipient SHOULD
treat it as type "application/octet-stream".

I know that it is unlikely to happen but I would stay away for `!!`. Maybe if it is not there, we fall back to the `System.currentTimeMillis()` ?

huiyiqun:
I have remove the `!!`.

kittinunf:
OK looks great. I'll merge this!

PR #142, opened 2017-03-15
FuelError.toString provides more useful information

mammothbane:
I was chasing down a bug in an Android application that was resulting in a failure, but the exception that caused it didn't provide a `message`, so I was staring at a bunch of `Failure { Exception : null }` in my logcat, which wasn't very helpful.

This change causes `FuelError.toString` to provide the fully-qualified name of the causing exception and its message (if present), defaulting to `"<no message>"` otherwise (since reading `null` if you're not expecting an error message in the first place isn't very helpful).


kittinunf:
Arh. Thank you so much. I will merge this in.

Issue #143, opened 2017-03-19
Android block mode -> Cannon resolve responseString


Tchoupinax:
Hello, 

I have a problem with trying to use the blocking mode code.
```Java
try {
    Triple<Request, Response, String> data = Fuel.get("http://www.google.com").responseString();
    Request request = data.getFirst();
    Response response = data.getSecond();
    Result<String,FuelError> text = data.getThird();
} 
catch (Exception networkError) {
}
```

First of all, it seems to be good to replace _Triple<Request, Response, String>_ by _Triple<Request, Response,  **Result<String,FuelError>**>_ (Android studio helps me to find it. ^^)

However for this line :
```Java
Triple<Request, Response, String> data = Fuel.get("http://www.google.com").responseString();
```

responseString() triggers "cannot resolve methode responseString()". I tried to change the variable type, try many things but don't know why it does not work. response() works seemingly but i want my string ! :D

Anyone has the solution ? 
Thanks for your time.

dyas90:
Same problem with me. 
i use this code
`val (request, response, result) = "http://httpbin.org/get".httpGet().responseString()`
get result `[Failure: Exception : null]`
but if i use code
`"http://httpbin.org/get".httpGet().responseJson {req, resp, res -> }`
get result `[Success: com.github.kittinunf.fuel.android.core.Json@4c0076a]`
what wrong?

kittinunf:
Hi, thank you you both for reporting!

The full explanation of the exception is this (Sorry, I think the current version the exception string is not good enough, it doesn't provide useful information).

```
java.lang.AssertionError: 
Expected: null
     but: was <Exception : sun.security.validator.ValidatorException: PKIX path building failed: java.security.cert.CertPathBuilderException: No issuer certificate for certificate in certification path found.>
Expected :null
     
Actual   :<Exception : sun.security.validator.ValidatorException: PKIX path building failed: java.security.cert.CertPathBuilderException: No issuer certificate for certificate in certification path found.>
```

So, it looks like you need to do either;

1. Configure the SSLContext with a TrustManager that accepts any cert or
2. Configure the SSLContext with an appropriate trust store that includes your cert
3. Add the cert for that site to the default java trust store.

You should be able to configure those via [FuelManager](https://github.com/kittinunf/Fuel/blob/master/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/FuelManager.kt)

Tchoupinax:
Hello,
[...]

PR #145, opened 2017-03-22
Add Test to show potential cert problem

kittinunf:
### What's in this PR?

Just add small test to address issue #143. In this test, we use `TrustManager` that accepts anything, so it is not advisable to use in real app.


kittinunf:
Merging this in!

PR #156, opened 2017-05-05
Add PATCH method for Fuel

### LINKED ISSUE #133: Missing PATCH-Verb
Is there a possibility to issue PATCH-requests without using the semi-standard `X-HTTP-Method-Override` header, e.g. `Fuel.patch(...)`?

Maybe I'm just not seeing it, there is a patch method defined [here](https://github.com/kittinunf/Fuel/blob/5ce5955b34a2f237ed3b3c223415da74619ad4ab/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Method.kt#L8) but it's not exposed like GET, POST, PUT or DELETE. Is there a reason for this?

kittinunf:
### What's in this PR?

This PR mitigates the issue where Fuel don't expose PATCH method for client to use. This PR will remedy that.

Merging this should close #133

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.0'
+        classpath 'com.android.tools.build:gradle:2.3.1'
         classpath 'com.dicedmelon.gradle:jacoco-android:0.1.1'
 
         classpath 'com.novoda:bintray-release:0.3.4'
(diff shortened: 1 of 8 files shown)


babedev:
Something wrong with tests?

kittinunf:
It seems like httpbin.org doesn't support POST method for httpbin.org/patch endpoint.

kittinunf:
🎉 🎉 🎉

babedev:
Wow, let's merge this.

PR #158, opened 2017-05-06
Multiple uploads example

babedev:
### What's in this PR?

Introducing multiple files upload with example.


kittinunf:
LGTM 🚢 🇮🇹

Issue #161, opened 2017-05-08
Customizations for multi file upload


Querschlag:
First of all, great thing that multi file upload has been implemented!

I propose adding some options to allow setting
- media type
- file name
- field value

for example: in OkHttp I can do this:
```
body.addFormDataPart("images[]", "image_name.jpg", RequestBody.create(MediaType.parse("image/jpeg"), imageFile))
```

babedev:
Sound interesting. It will be supported in next release.

o-nix:
> media type
Related to #155

Querschlag:
Brilliant!
Could you also add an option for the field value the files get assigned to?
Maybe a data class that hold the files + the field value...

yoavst:
@Querschlag I'm not knowledgeable in the field of HTTP Upload, so was the feature you asked for was added?

kittinunf:
I think it should be supported already. Closing this.

PR #162, opened 2017-05-08
Custom form data part

### LINKED ISSUE #155: Externalize `Content-Type` detection for `multipart/form-data`
Currently it uses `URLConnection.guessContentTypeFromName(file.name)`, so my `certificate.pem` becomes `Content-Type: null`.

babedev:
### What's in this PR?

Let user input specific file's name and content type using `DataPart`

Example

```Kotlin
.upload("/post")
.dataParts { request, url ->
   listOf(
      DataPart(File(currentDir, "lorem_ipsum_short.tmp"), "image/jpeg"),
      DataPart(File(currentDir, "lorem_ipsum_long.tmp"), "second-file", "image/jpeg")
   )
}
.responseString { req, res, result -> }
```

And also avoid `null` content type by using `MimetypesFileTypeMap`

Merging this should close #155 and #161

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/DataPart.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/DataPart.kt
@@ -1,0 +1,8 @@
+package com.github.kittinunf.fuel.core
+
+import java.io.File
+
+data class DataPart(
+        var fileName: String = "",
+        var mediaType: String = "",
+        var file: File)
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/DataPart.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/DataPart.kt
@@ -3,6 +3,6 @@ package com.github.kittinunf.fuel.core
 import java.io.File
 
 data class DataPart(
-        var fileName: String = "",
-        var mediaType: String = "",
-        var file: File)
+        val fileName: String = "",
+        val mediaType: String = "",
+        val file: File)
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/DataPart.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/DataPart.kt
@@ -3,6 +3,6 @@ package com.github.kittinunf.fuel.core
 import java.io.File
 
 data class DataPart(
-        val fileName: String = "",
-        val mediaType: String = "",
-        val file: File)
+        val file: File,
+        val name: String = file.name.split(".").getOrElse(0) { "" },
+        val type: String = "")
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
@@ -45,6 +45,9 @@ class Request : Fuel.RequestConvertible {
 
     var name = ""
 
+    var names = arrayListOf<String>()
+    var mediaTypes = arrayListOf<String>()
+
     //underlying task request
     val taskRequest: TaskRequest by lazy {
         when (type) {
@@ -131,11 +134,34 @@ class Request : Fuel.RequestConvertible {
         return this
     }
 
-    fun sources(source: (Request, URL) -> Iterable<File>): Request {
+    fun dataParts(dataParts: (Request, URL) -> Iterable<DataPart>): Request {
+        val parts = dataParts.invoke(request, request.url)
+
+        mediaTypes.clear()
+        mediaTypes.addAll(parts.map { it.mediaType })
+
+        names.clear()
+        names.addAll(parts.map { it.fileName })
+
         val uploadTaskRequest = taskRequest as? UploadTaskRequest ?: throw IllegalStateException("source is only used with RequestType.UPLOAD")
 
         uploadTaskRequest.apply {
-            sourceCallback = source
+            sourceCallback = { request, url ->
+                parts.map { it.file }
+            }
+        }
+
+        return this
+    }
+
+    fun sources(sources: (Request, URL) -> Iterable<File>): Request {
+        mediaTypes.clear()
+        names.clear()
+
+        val uploadTaskRequest = taskRequest as? UploadTaskRequest ?: throw IllegalStateException("source is only used with RequestType.UPLOAD")
+
+        uploadTaskRequest.apply {
+            sourceCallback = sources
         }
         return this
     }
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
@@ -45,8 +45,8 @@ class Request : Fuel.RequestConvertible {
 
     var name = ""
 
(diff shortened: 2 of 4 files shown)


fuel/src/main/kotlin/com/github/kittinunf/fuel/core/DataPart.kt
```diff
@@ -0,0 +1,8 @@
+package com.github.kittinunf.fuel.core
+
+import java.io.File
+
+data class DataPart(
+        var fileName: String = "",
```
kittinunf: Prefer to use `val` if possible

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/DataPart.kt
```diff
@@ -0,0 +1,8 @@
...
+        var mediaType: String = "",
```
kittinunf: Here as well

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
```diff
@@ -45,6 +45,9 @@ class Request : Fuel.RequestConvertible {
 
     var name = ""
 
+    var names = arrayListOf<String>()
```
kittinunf: Prefer kotlin type either `List` or `MutableList`

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
```diff
@@ -45,6 +45,9 @@ class Request : Fuel.RequestConvertible {
...
+    var mediaTypes = arrayListOf<String>()
```
kittinunf: Here as well

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
```diff
@@ -131,11 +134,34 @@ class Request : Fuel.RequestConvertible {
...
+    fun dataParts(dataParts: (Request, URL) -> Iterable<DataPart>): Request {
+        val parts = dataParts.invoke(request, request.url)
+
+        mediaTypes.clear()
+        mediaTypes.addAll(parts.map { it.mediaType })
+
+        names.clear()
+        names.addAll(parts.map { it.fileName })
+
+        val uploadTaskRequest = taskRequest as? UploadTaskRequest ?: throw IllegalStateException("source is only used with RequestType.UPLOAD")
```
kittinunf: You might wanna move this to the top as to create a golden path

fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestUploadTest.kt
```diff
@@ -235,4 +231,45 @@ class RequestUploadTest : BaseTestCase() {
...
+        var error: FuelError? = null
+
+        manager.upload("/post", param = listOf("foo" to "bar"))
+                .sources { request, url ->
+                    listOf(File(currentDir, "lorem_ipsum_short.tmp"),
+                            File(currentDir, "lorem_ipsum_long.tmp"))
+                }
+                // Override with Data parts
+                .dataParts { request, url ->
+                    listOf(DataPart("first-file", "image/jpeg", File(currentDir, "lorem_ipsum_short.tmp")),
```
kittinunf: We do have to specify `File` twice ? Once in `source` and `dataPart` block. Seem a little inconvenient... Wonder we can design an API to simplify this?

babedev: Oh, actually only `dataParts` is fine. Example show what if we call `dataParts` it will override all files in `sources`.

kittinunf: Ahhh ... I see. 🙆  . But do you think our client would be confused by them?

babedev: Um... may be. I will remove `sources` in this test case. :)

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
```diff
@@ -45,6 +45,9 @@ class Request : Fuel.RequestConvertible {
 
     var name = ""
 
+    var names = mutableListOf<String>()
```
kittinunf: This should be `val`

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
```diff
@@ -45,6 +45,9 @@ class Request : Fuel.RequestConvertible {
...
+    var mediaTypes = mutableListOf<String>()
```
kittinunf: This too :)

fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestUploadTest.kt
kittinunf: And I suggest that we have one overloading method. in the case, where user just want to upload one file, so it should be extremely simple.

Like
``` Kotlin
Fuel.upload("/upload").source { req, url -> File(currentDir, "foo.txt") }.responseString()
```

Maybe we can construct DataPart object from a File object. For example, if the File is supplied, we can infer `name` by using the filename.

kittinunf:
*copy here

And I suggest that we have one overloading method. in the case, where user just want to upload one file, so it should be extremely simple.

Like
``` Kotlin
Fuel.upload("/upload").source { req, url -> File(currentDir, "foo.txt") }.responseString()
```

Maybe we can construct `DataPart` object from a `File` object. With Kotlin, I think you can do `DataPart(val file: File, val name: String = file.name, val type: String? = null)`. 

Hence, if the `File` is supplied, we can infer `name` by using the filename.

So, in that case, we can keep the `source` API and use that to upload a single file only? What do you think?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
```diff
@@ -134,15 +134,23 @@ class Request : Fuel.RequestConvertible {
         return this
     }
 
+    fun dataPart(dataPart: (Request, URL) -> DataPart): Request {
```
babedev: @kittinunf `dataPart` for single file as you suggested. :D

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
```diff
@@ -131,11 +134,39 @@ class Request : Fuel.RequestConvertible {
         return this
     }
 
-    fun sources(source: (Request, URL) -> Iterable<File>): Request {
+    fun dataPart(dataPart: (Request, URL) -> DataPart): Request {
```
kittinunf: ah sorry for not being more clear ... I mean this to become `fun source(source: (Request, URL) -> File)`. So it is easy for usage with only one `File`

kittinunf:
:claps: :green:

kittinunf:
@babedev merge?

babedev:
Merged :D

PR #164, opened 2017-05-09
Migrate to RxJava2 and kotlin 1.1.2

Judrummer:
let's use RxJava2 :)

### DIFF
--- a/fuel-rxjava/src/main/kotlin/com/github/kittinunf/fuel/rx/RxFuel.kt
+++ b/fuel-rxjava/src/main/kotlin/com/github/kittinunf/fuel/rx/RxFuel.kt
@@ -15,9 +15,8 @@ fun <T : Any> Request.rx_responseObject(deserializable: Deserializable<T>) = rx_
 
 private fun <T : Any> Request.rx_response(deserializable: Deserializable<T>): Single<Pair<Response, Result<T, FuelError>>> =
         Single.create { emitter ->
-            response(deserializable) { _, response, result ->
-                emitter.onSuccess(response to result)
-            }
+            val (_, response, result) = response(deserializable)
+            emitter.onSuccess(response to result)
             emitter.setCancellable { this.cancel() }
         }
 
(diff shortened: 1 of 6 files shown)


kittinunf:
👏 👏 👏

fuel-rxjava/src/main/kotlin/com/github/kittinunf/fuel/rx/RxFuel.kt
```diff
@@ -24,15 +24,15 @@ private fun <T : Any> Request.rx_response(deserializable: Deserializable<T>): Ob
                 when (result) {
```
kittinunf: I think in RxJava2, there is `Observable.create` for us to use it safely. Would you mind the whole implementation to use `.create` instead?

fuel-rxjava/src/main/kotlin/com/github/kittinunf/fuel/rx/RxFuel.kt
```diff
@@ -24,15 +24,15 @@ private fun <T : Any> Request.rx_response(deserializable: Deserializable<T>): Ob
...
+                        source.onComplete()
                     }
 
                     is Result.Failure -> {
                         source.onError(result.error)
                     }
                 }
             }
-            source.doOnUnsubscribe { this.cancel() }
+            source.doOnDispose { this.cancel() }
```
kittinunf: In here, it should be `source.setCancellable`

fuel-rxjava/src/main/kotlin/com/github/kittinunf/fuel/rx/RxFuel.kt
Judrummer: Do you mean this? @kittinunf 

```
private fun <T : Any> Request.rx_response(deserializable: Deserializable<T>): Observable<Pair<Response, T>> =
        Observable.create { emitter ->
            response(deserializable) { request, response, result ->
                when (result) {
                    is Result.Success -> {
                        emitter.onNext(response to result.value)
                        emitter.onComplete()
                    }

                    is Result.Failure -> {
                        emitter.onError(result.error)
                    }
                }
            }
            emitter.setCancellable { this.cancel() }
        }
```

kittinunf: Aha. I would love to use this opportunity not using `.onError`, but using only `onNext`. Also, I think it might be just better to use `Single<Pair<Response, T>>`. What do you think?

kittinunf:
Please consider my proposal 🙏

Judrummer:
okayy

kittinunf:
🍏

PR #171, opened 2017-05-23
Fix typo in function usage.

kennycason:


babedev:
It is ok to use 
`"http://www.example.com/user/1".httpGet().responseObject(User.Deserializer()) { req, res, result ->
}`

if the last parameter is lambda.

kennycason:
@babedev Got it. Then I was likely doing something wrong because I could not get it to compile.

PR #172, opened 2017-05-26
Added ability to send/receive big files by using real stream instead of bytestream.

savardd:
For download, the destination task request can now use the read stream from the request, permitting to save a big file to disk.
Also, when deserializing with the two 'stream' overload, the real stream is accessible, permitting to deserialize big objects if needed.  But, all code compatibility is kept and the stream is not accessible outside the request overload or deserialization object.

For upload, using the http stream permit to send big files to server.  The call back is called twice, once without the stream object (to retrieve the expected size), the second time with the stream object, for the callback to write in it.  Also, the computed size from the first call is also specified to facilitate progress code.


kittinunf:
@savardd Thank you so much. Let me take a look and get back to you!

kittinunf:
i think it looks really good. Can you add test downloading hug files as well? After that, this looks good to merge.

savardd:
I just added two new tests, one for big file download and another to test the head request with gzip compression enabled.

kittinunf:
Awesome!! Thank you, I will merge this in!

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Response.kt
```diff
@@ -12,7 +16,14 @@ class Response {
     var httpContentLength = 0L
 
     //data
-    var data = ByteArray(0)
+    internal var dataStream: InputStream = ByteArrayInputStream(ByteArray(0))
```
FrederikNS: Is it intentional that `dataStream` is `internal`, and that `data` was switched from `var` to `val`?

This makes it really hard to override the built in HttpClient, and therefore also very hard to stub out the actual web requests when unittesting.

If `dataStream` was not `internal`, or `data` was a `var`, I would be able to solve #186

kittinunf: Ah .. I might have overlooked that usecase! I think your point is valid indeed. I will open a patch on this for v 1.8.1. 😓  really sorry for this!

PR #175, opened 2017-06-01
LiveData module

KucherenkoIhor:
```
+ fun httpLiveDataSupport() {
+        "http://jsonplaceholder.typicode.com/photos/1".httpGet().liveData_object(Photo.Deserializer())
+                .observeForever { result ->
+                    Log.d(TAG, result.toString())
+                }
+    }`


kittinunf:
👍  Thank you so much for this!!! Btw, can you please add a bit of tests for us? And also, I wanna ask your honest feel about the naming. Do you think the `_` thing is a bit weird? (I know that I have used it in the past and I think you just probably want to follow the convention). 

I think this is a brand new APIs so we might be able to come up with something if it is clearly a better name.

KucherenkoIhor:
Hi. I think we can just remove _ . Which tests do you mean? AndroidTest? or
just samples of usage?

kittinunf:
Thanks, I am merging this in.

Issue #177, opened 2017-06-02
HTTP HEAD tries to read gzip input stream and throws an exception


seanmadden:
Running

`curl -i -k -I  -H "Accept-Encoding:compress;q=0.5, gzip;q=1.0" "https://www.google.com"`

results in 
```
HTTP/1.1 200 OK
Date: Fri, 02 Jun 2017 21:04:59 GMT
Expires: -1
Cache-Control: private, max-age=0
Content-Type: text/html; charset=ISO-8859-1
P3P: CP="This is not a P3P policy! See https://www.google.com/support/accounts/answer/151657?hl=en for more info."
Content-Encoding: gzip
Server: gws
X-XSS-Protection: 1; mode=block
X-Frame-Options: SAMEORIGIN
Set-Cookie: NID=104=k8idQPM6UGdqGRBaYCz_G5cU6UbsH1Fnl3IA_9VuGF_RaGcG6DiHrxhRdL5AwCx4k0Yn-vezV-y_VvzG9tPE4Aj4U0orMw748TJV1DuXJHnt0OMIpmE1yGCQkv8Fcp7KtFQr4xXwK3LMQ89X; expires=Sat, 02-Dec-2017 21:04:59 GMT; path=/; domain=.google.com; HttpOnly
Alt-Svc: quic=":443"; ma=2592000; v="38,37,36,35"
Transfer-Encoding: chunked
```

Which has **Content-Encoding: gzip** - this causes Fuel to try to parse an empty GZIP input stream (since HEAD requests have no bodies) causing an EOF exception to be thrown.

I'm submitting a pull request shortly.

kittinunf:
Can we close this down? 👍

seanmadden:
Yep 👍

PR #178, opened 2017-06-02
Stop trying to read http head response input streams.

seanmadden:
This fixes issue #177 

From RFC2619 (https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html)

The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response. The metainformation contained in the HTTP headers in response to a HEAD request SHOULD be identical to the information sent in response to a GET request. This method can be used for obtaining metainformation about the entity implied by the request without transferring the entity-body itself. This method is often used for testing hypertext links for validity, accessibility, and recent modification.

So, it is valid to return the Content-Encoding: gzip header, but we must not try to read the empty stream as a gzip stream.


kittinunf:
Thank you so much for this. I will be merging this in.

PR #190, opened 2017-07-04
Feature/gson module

KucherenkoIhor:
Hi)
To avoid creation boilerplate deserializer I added gson module support.
If user want to work with Gson out of the box, it can just add dependencies.
And after this you can just specify type of your model and you don't need to create Deserializer class.
```kotlin
 "http://jsonplaceholder.typicode.com/photos/1".httpGet()
                .responseObject { request: Request, response: Response, result: Result<Photo, FuelError> ->
                    Log.d(TAG, request.toString())
                    updateUI(response, result)
                }
```


fuel-gson/build.gradle
```diff
@@ -0,0 +1,85 @@
...
+}
+
+jacocoAndroidUnitTestReport {
+    csv.enabled false
+    html.enabled true
+    xml.enabled true
+}
+
+dependencies {
+    compile "android.arch.lifecycle:extensions:$rootProject.archLifecycleVersion"
```
kittinunf: Why do we need this?

fuel-gson/build.gradle
```diff
@@ -0,0 +1,85 @@
...
+
+apply plugin: 'com.novoda.bintray-release'
+
+apply plugin: 'jacoco-android'
+
+repositories {
+    jcenter()
+}
+
+android {
```
kittinunf: What do you think about making this module Android-free? I think that Gson + Fuel could target to JVM itself.

kittinunf:
Looks awesome to me! I am merging this in.

Issue #150, opened 2017-04-24
Could not initialize class com.github.kittinunf.fuel.core.FuelManager


kuy:
I got errors when I call Fuel API but I couldn't realize what's wrong.
My question: Does Fuel support Kotlin 1.1.x?

### Environment

JVM: 1.8
Kotlin: 1.1
Fuel: 1.5.0
Result: 1.0.5

### Stack Trace

```
2017-04-24 20:24:54,497 [http-nio-8080-exec-3] ERROR org.apache.catalina.core.ContainerBase.[Tomcat].[localhost].[/].[servletContainer] - Servlet.service() for servlet [servletContainer] in context with path [] threw exception [org.glassfish.jersey.server.ContainerException: java.lang.NoClassDefFoundError: Could not initialize class com.github.kittinunf.fuel.core.FuelManager] with root cause
java.lang.NoClassDefFoundError: Could not initialize class com.github.kittinunf.fuel.core.FuelManager
	at com.github.kittinunf.fuel.Fuel$Companion.request(Fuel.kt:108)
	at com.github.kittinunf.fuel.Fuel$Companion.get(Fuel.kt:32)
	at com.github.kittinunf.fuel.FuelKt.httpGet(Fuel.kt:126)
	at com.github.kittinunf.fuel.FuelKt.httpGet$default(Fuel.kt:125)
	at jp.entrepedia.contract.ContractResources.sendMail(ContractResources.kt:60)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.glassfish.jersey.server.model.internal.ResourceMethodInvocationHandlerFactory$1.invoke(ResourceMethodInvocationHandlerFactory.java:81)
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher$1.run(AbstractJavaResourceMethodDispatcher.java:144)
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.invoke(AbstractJavaResourceMethodDispatcher.java:161)
[...]

PR #194, opened 2017-07-13
Fix for #150

ParkerK:
https://github.com/kittinunf/Fuel/issues/150 - javax.activation is only available in the JavaSDK, it's not in the Android core library


kittinunf:
Thank you so much for this :) I will merge this in.

ParkerK:
Happy to help, thanks for the library :)

PR #203, opened 2017-07-23
Implement uploading data from InputStreams

### LINKED ISSUE #176: File upload from InputStream directly
I try to upload file from internal Uri using ContentResolver. I have got filename and Uri (I can open inputstream over ContentResolver) but I have no File. 
There's a stream and a filename inside UploadTaskRequest but I cannot access them. 
The only workaround I found is to create temporary file.
Could you please consider about create alternative way of UploadTaskRequest initialization?

tsudoko:
Fixes #176.

Unfortunately there's no way to make `Blob`s compatible with existing `DataPart`s without breaking backwards compatibility due to the public `file` property.

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,6 +347,14 @@ Fuel.upload("/post").sources { request, url ->
 }
 ```
 
+### Upload from an `InputStream`
+
+``` Kotlin
+Fuel.upload("/post").blob { request, url ->
+    Blob("filename.png", 1166976, { someObject.getInputStream() })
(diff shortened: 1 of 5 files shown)


kittinunf:
This looks awesome! Would you mind adding some doc in the README.md file to inform users about this? Aside from that, I would be happy to merge this.

kittinunf:
Can't say thank you enough for this! Lovely 👍

PR #208, opened 2017-08-06
Add HTTP status code 303 as valid redirect

ParkerK:
Currently Fuel handles HTTP Status Code 303 as an error. It should be handled as a redirect though: 
https://en.wikipedia.org/wiki/HTTP_303
https://httpstatuses.com/303


tsudoko:
It might be a good idea to make all 3xx responses be treated like that to avoid further misses (this is the second PR of that kind) and ensure the library handles new, not yet introduced 3xx responses properly.

RFC 7231, section [6.4][]:
```
   The 3xx (Redirection) class of status code indicates that further
   action needs to be taken by the user agent in order to fulfill the
   request.  If a Location header field (Section 7.1.2) is provided, the
   user agent MAY automatically redirect its request to the URI
   referenced by the Location field value, even if the specific status
   code is not understood.
```

[6.4]: https://tools.ietf.org/html/rfc7231#section-6.4

sealemar:
Not all **HTTP 3xx** are redirects --- see [10.3 Redirection 3xx](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html)

> **10.3 Redirection 3xx**
This class of status code indicates that **further action** needs to be taken by the user agent in order to fulfill the request...

Note the **further action** part above.

For example, the [304 NOT MODIFIED](https://httpstatuses.com/304):

> A conditional GET or HEAD request has been received and would have resulted in a 200 OK response if it were not for the fact that the condition evaluated to false.

> The server generating a 304 response MUST generate any of the following header fields that would have been sent in a 200 OK response to the same request: Cache-Control, Content-Location, Date, ETag, Expires, and Vary.

In other words, the other 30x use **Location** field to target the redirect, the 304 does not use the **Location** field.

tsudoko:
My bad, I erroneously assumed the library just doesn't redirect if there's no `Location` header.

sealemar:
The **further action** part of the specification is ambiguous --- so strict interpretations must be avoided. It is unknown how the specification will be extended in future. Only unambiguous parts of 3xx redirects should be treated as such.

ParkerK:
Would it be better then to check if the status code is 3xx && has a location value in the header? I assume that if it has a location value, it's fair to say it's a redirect status

tsudoko:
That's actually what's suggested in RFC 7231.

sealemar:
The [Location section of RFC 7231](https://tools.ietf.org/html/rfc7231#section-7.1.2) does not require a redirect.

The [6.4. Redirection 3xx](https://tools.ietf.org/html/rfc7231#section-6.4) says:

> If a Location header field (Section 7.1.2) is provided, the user agent MAY automatically redirect its request to the URI referenced by the Location field value, even if the specific status code is not understood. Automatic redirection needs to done with care for methods not known to be safe, as defined in Section 4.2.1, since the user might not wish to redirect an unsafe request.

Note the **unsafe request**.

[Section 4.2.1 Safe Methods](https://tools.ietf.org/html/rfc7231#section-4.2.1)

> Request methods are considered "safe" if their defined semantics are essentially read-only; i.e., the client does not request, and does not expect, any state change on the origin server as a result of applying a safe method to a target resource. Likewise, reasonable use of a safe method is not expected to cause any harm, loss of property, or unusual burden on the origin server.

> Of the request methods defined by this specification, the GET, HEAD, OPTIONS, and TRACE methods are defined to be safe.

I.e. **6.4.3. 302 Found** of the same RFC MAY NOT be safe, since:

> Note: For historical reasons, a user agent MAY change the request method from POST to GET for the subsequent request.  If this behavior is undesired, the 307 (Temporary Redirect) status code    can be used instead.

Note **MAY** word.

Let's go back to [6.4. Redirection 3xx](https://tools.ietf.org/html/rfc7231#section-6.4):

> ...If a Location header field (Section 7.1.2) is provided, the user agent MAY automatically redirect its request to the URI referenced by the Location field value...

As per [1.1. Conformance and Error Handling](https://tools.ietf.org/html/rfc7231#section-1.1), the word MAY is defined in [rfc2119](https://tools.ietf.org/html/rfc2119):

> 5. MAY   This word, or the adjective "OPTIONAL", mean that an item is truly optional...

The specification defines a redirect for 3xx error codes based solely on **Location** header _(even if the specific status code is not understood)_ as **truly optional**.

That MAY lead to a situation when the RFC is extended for additional status code with a vague definition and a server-side is implemented with a conditional redirect, where the **redirect to** resource is defined in **Location** header and the condition is defined as per [5.2. Conditionals](https://tools.ietf.org/html/rfc7231#section-5.2).

With that said it would be safer not to do a blanket redirect based solely on **Location** field.

sealemar:
Another thing that I've noticed in the RFC 7231 is #214

kittinunf:
Thanks! @ParkerK 

Do you mind to add some tests to verify this? Thanks 👍

ParkerK:
Sure thing, I'll update when they're done

ParkerK:
Sorry, forgot about this PR. What tests should I add? I'm kind of clueless when it comes to tests.

ParkerK:
@kittinunf sorry, I forgot about this! I added a test, I haven't really written tests before, so I just looked at the existing ones you have. Please let me know if I did anything wrong!

ParkerK:
It looks like @kindraywind added InterceptorTest.kt recently as well

ParkerK:
@kittinunf - The Travis CI build failed due to a faulty test. I submitted a separate PR to fix the test: https://github.com/kittinunf/Fuel/pull/354

ParkerK:
@kittinunf would you be able to approve the merge of master into this patch?

SleeplessByte:
If you want this implementation to work how browsers work, the implementation is incorrect:
307 and 308 ' behaviour is correct: the request method for the redirected request.
301 and 303 ' behaviour is incorrect: the request method should be changed to `GET` (307 and 308 were introduced to differ from this)
302 ' behaviour is dubious: should have been like 307 but is implemented as 301, so you probably want to implement it like 301.

kittinunf:
Sorry for my ignorance for not taking at this.

Let's try to merge this in I think we can merge this first and improve for 301 and 303 in the different PR? @ParkerK do you think we should have tests for other cases as well? After that, we can merge this in? 👍

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/RedirectionInterceptor.kt
```diff
@@ -2,39 +2,51 @@ package com.github.kittinunf.fuel.core.interceptors
 
 import com.github.kittinunf.fuel.core.*
 import java.net.MalformedURLException
+import java.net.URI
 import java.net.URL
-import javax.net.ssl.HttpsURLConnection
+import javax.net.ssl.HttpsURLConnection.HTTP_MULT_CHOICE //300
```
kittinunf: [Nits] I dont think we need these comment

SleeplessByte:
Sure; but since this PR is supposed to add `303`, I wouldn't want that behaviour to be incorrect.

https://github.com/kittinunf/Fuel/pull/208/files#diff-78105644d5979bb85e0cae5ee1f5fc98R27 should be 'GET' for 301 and 303, and if you want to match browsers also 302. The rest can stay the same 🥇

kittinunf:
Superseded by #388

Issue #198, opened 2017-07-19
Condition on README snippet seems inverted


jivimberg:
Error condition on README async example for Kotlin seems to be inverted: https://github.com/kittinunf/Fuel/blob/62daa920d93fc1d7cc5ef1bd3dbbd7b6804c475c/README.md#L78

kittinunf:
Can you elaborate a bit on that? Even better, if you can open a PR to resolve that.

kittinunf:
Close this.

PR #210, opened 2017-08-08
Inverting condition on example

jivimberg:
Inverting error handling condition on README.md example to fix issue #198


kittinunf:
Nice! Thank you 👏

Issue #217, opened 2017-08-16
Iterating response headers throws NPE while the same works in OkHttp


asarkar:
```
class Main {
    private val url = "https://fantasy.premierleague.com/drf/elements/"

    fun syncGetOkHttp() {
        println("\n===")
        println("OkHttp")
        println("===")
        val client = OkHttpClient()

        val request = Request.Builder()
                .url(url)
                .header("Accept", "application/json")
                .build()

        println("Request headers ===>")
        request.headers().toMultimap().forEach { k, v -> println("$k ===> $v.") }

        val response = client.newCall(request).execute()
        if (!response.isSuccessful) throw IOException("Unexpected code: ${response.code()}.")

        println("Status code ===> ${response.code()}.")
        println("Body ===> ${String(response.body()!!.bytes())}.")

        val responseHeaders = response.headers()

        println("Response headers ===>")
        responseHeaders.toMultimap().forEach { k, v -> println("$k ===> $v.") }
    }

    fun syncGetFuel() {
        println("\n===")
        println("Fuel")
        println("===")

        val (request, response, result) = url.httpGet().run {
            httpHeaders.put("Accept", "application/json")

            responseString()
        }

        println("Request headers ===>")
        request.httpHeaders.forEach { k, v -> println("$k ===> $v.") }

        println("Status code ===> ${response.httpStatusCode}.")

        when (result) {
            is Result.Failure -> {
                result.error.exception.printStackTrace()
            }
            is Result.Success -> {
                println("Body ===> ${result.get()}.")

                println("Response headers ===>")
                response.httpResponseHeaders.forEach { k, v -> println("$k ===> $v.") }
            }
        }
    }
}
```

Output:
===
OkHttp
===
Request headers ===>
accept ===> [application/json].
Status code ===> 200.
Body ===> . // blank! Issue reported in this [SO thread](https://stackoverflow.com/q/45702466/839733)
Response headers ===>
accept-ranges ===> [bytes].
connection ===> [close].
content-length ===> [0].
content-type ===> [application/json].
date ===> [Wed, 16 Aug 2017 04:00:44 GMT].
retry-after ===> [0].
server ===> [Varnish].
via ===> [1.1 varnish].
x-cache ===> [MISS].
x-cache-hits ===> [0].
x-served-by ===> [cache-sea1047-SEA].
x-timer ===> [S1502856045.965729,VS0,VE0].

Fuel
===
Request headers ===>
[...]

PR #218, opened 2017-08-17
Filter out null keys in headerFields map

ParkerK:
`URLConnection`s `getHeaderFields()` contains a null key for the HTTP response code. I have no idea why, and other libraries sort out null keys, so I added it here. It's not Fuel's fault, it's just the way `URLConnection works for some reason

Fixes issue #217


kittinunf:
Nice!

Issue #216, opened 2017-08-15
Router with Enum & Interface


matteocrippa:
hey @kittinunf first of all compliments for the huge work, I'm just approaching Kotlin on android and I feel this library is really interesting, even more than Retrofit.

I'm an iOS developer and I feel really comfortable in using `enum` for easily manage remote call, I'm trying in porting this solution to Fuel (already tested in a project) with something like

```
enum class FuelHttpMetod {
    get, post, put, patch, delete
}

interface FuelRouter {
    val basePath: String
    val method: FuelHttpMetod?
    val path: String?

    fun exec(params: List<Pair<String, Any?>>?, handler: (Result<Json, FuelError>) -> Unit) {
        // set base path
        FuelManager.instance.basePath = basePath
        // handle method
        method?.let { m ->
            var remote: Request?
            when(m) {
                FuelHttpMetod.get -> remote = path?.httpGet(params)
                FuelHttpMetod.post -> remote = path?.httpPost(params)
                else -> return
            }

            remote?.responseJson { _, _, result ->
                handler(result)
            }
        }
    }
}
```

this easily allows me to create an enum like

```
enum class WeatherApi: FuelRouter {

    getWeather;

    override val basePath = "https://www.metaweather.com"

    override val method: FuelHttpMetod?
        get() {
            when(this) {
                getWeather -> return FuelHttpMetod.get
                else -> return null
            }
        }

    override val path: String?
        get() {
            when(this) {
                getWeather -> return "/api/location/search/"
                else -> return null
            }
        }
}
```

and then use with

```
WeatherApi.getWeather.exec(params = listOf("query" to "london"), handler = { result ->
            Log.d("weatherApi", result.toString())
        })
```

if you think this can be interesting for you and Fuel, I can make a PR for this

matteocrippa:
I think this is even a little better https://gist.github.com/matteocrippa/9d6824f378202eca5886cb338eb7f0fc 

I moved using your native `Method` enums and improved the interface with new feature to better fit the common usage with REST APIs

kittinunf:
@matteocrippa First of all, thanks so much for those kind words. I am glad that you find `Fuel` useful.

Regarding to your comment on the concept of `Router`, actually, it is pretty interesting. I have a glimpse of how it could be, if you take a look of the interface there

[...]

PR #221, opened 2017-08-17
FuelRouting

matteocrippa:
As proposed in #216 here the related PR for the routing feature.

I have updated also the readme with the new feature and example (I also updated that with your suggestion using fold).


fuel/src/main/kotlin/com/github/kittinunf/fuel/util/FuelRouting.kt
```diff
@@ -0,0 +1,53 @@
...
+     */
+    override val request: Request
+        get() {
+            // generate the encoder according provided parameters, headers, path, etc.
+            val encoder = Encoding().apply {
+                this.baseUrlString = basePath
+                this.httpMethod = method
+                this.urlString = path
+                this.parameters = params
+                // FIXME: headers are missing
```
kittinunf: Is there anything we can do for this? What do you mean by `headers` are missing though?

matteocrippa: I see there's no way to set headers and inject them like parameters or so. Looking at `encoder` I don't see any particular way to append them to the request

yoavst: What's about appending them after you have a `Request` object?

matteocrippa: makes sense, I'll update in that way, I was expecting to set it in apply, but I totally miss that

kittinunf:
This looks great to me. If @yoavst say thumb up for this, I am happy to merge this in.

yoavst:
:+1:

Issue #220, opened 2017-08-17
HTTP header access is inconsistent between request and response


asarkar:
The request headers are accessed with `httpHeaders` while the response headers are accessed with `httpResponseHeaders`. Apart from being inconsistent, the prefix `http` seems unnecessary; fuel is a HTTP library. What other kinds of headers can it possibly return?

PR #223, opened 2017-08-20
Refactoring

yoavst:
When I tried (and failed) to convert Fuel to a multi-platform project, I've noticed that Fuel codebase needs a refactoring for two goals: 
* Make the code more Kotlin (primary constructors instead of `lateinit`s, use new methods from 1.1 instead of calling Java...)
* Hide Java implementation details out of the public interface

This refactoring attempt tries to hide the Java impl details when it is possible to do it without losing features. However the library is still Java library, and more step has to be taken to make it possible to use Fuel in multi platforms.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,7 @@ allprojects {
         publish_version = '1.9.0'
 
         //dependencies version
-        kotlin_version = '1.1.3'
+        kotlin_version = '1.1.4'
         result_version = '1.1.0'
         rxjava_version = '2.1.0'
         archLifecycleVersion = "1.0.0-alpha3"
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,7 @@ allprojects {
         publish_version = '1.9.0'
 
         //dependencies version
-        kotlin_version = '1.1.4'
+        kotlin_version = '1.1.4-3'
         result_version = '1.1.0'
         rxjava_version = '2.1.0'
         archLifecycleVersion = "1.0.0-alpha3"
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
@@ -1,7 +1,6 @@
 package com.github.kittinunf.fuel.core
 
 import com.github.kittinunf.fuel.Fuel
-import com.github.kittinunf.fuel.util.toHexString
 import java.net.MalformedURLException
 import java.net.URI
 import java.net.URISyntaxException
@@ -29,7 +28,7 @@ class Encoding(val httpMethod: Method,
                 modifiedPath += (querySign + queryParamString)
             }
             requestType == Request.Type.UPLOAD -> {
-                val boundary = System.currentTimeMillis().toHexString()
+                val boundary = System.currentTimeMillis().toString(16)
                 headerPairs += "Content-Type" to "multipart/form-data; boundary=" + boundary
             }
             else -> {
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
@@ -36,12 +36,13 @@ class Encoding(val httpMethod: Method,
                 data = queryFromParameters(parameters)
             }
         }
-        Request().apply {
-            httpMethod = method
-            this.path = modifiedPath
-            this.url = createUrl(modifiedPath)
-            this.type = requestType
-            this.parameters = parameters ?: emptyList()
+        Request(
+                httpMethod = method,
+                path = modifiedPath,
+                url = createUrl(modifiedPath),
+                type = requestType,
+                parameters = parameters ?: emptyList()
+        ).apply {
             header(headerPairs, false)
             if (data != null) body(data ?: "")
         }
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
@@ -37,7 +37,7 @@ class Encoding(val httpMethod: Method,
             }
         }
         Request(
-                httpMethod = method,
+                method = method,
                 path = modifiedPath,
                 url = createUrl(modifiedPath),
                 type = requestType,
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
@@ -15,7 +15,7 @@ class Encoding(val httpMethod: Method,
     private val encoder: (Method, String, List<Pair<String, Any?>>?) -> Request = { method, path, parameters ->
         var modifiedPath = path
         var data: String? = null
-        val headerPairs: MutableMap<String, Any> = defaultHeaders.toMutableMap()
+        val headerPairs = defaultHeaders.toMutableMap()
         when {
             encodeParameterInUrl(method) -> {
                 var querySign = ""
@@ -44,7 +44,7 @@ class Encoding(val httpMethod: Method,
                 parameters = parameters ?: emptyList()
         ).apply {
             header(headerPairs, false)
-            if (data != null) body(data ?: "")
+            if (data != null) body(data!!)
         }
 
     }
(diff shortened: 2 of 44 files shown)


fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
```diff
@@ -65,23 +66,15 @@ class Encoding : Fuel.RequestConvertible {
...
-
-    companion object {
-        private fun defaultHeaders(): MutableMap<String, Any> = mutableMapOf("Accept-Encoding" to "compress;q=0.5, gzip;q=1.0")
-    }
+    private fun queryFromParameters(params: List<Pair<String, Any?>>?): String = params.orEmpty()
+            .filterNot { it.second == null }
+            .mapTo(mutableListOf()) { "${it.first}=${it.second}" }
+            .joinToString("&")
 
+    private fun defaultHeaders(): MutableMap<String, Any> = mutableMapOf("Accept-Encoding" to "compress;q=0.5, gzip;q=1.0")
```
kittinunf: Having return as `MutableMap` might not be such a great idea. Can we use just `Map` and let the client convert to `mutable` counterpart themselves?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
```diff
@@ -65,23 +66,15 @@ class Encoding : Fuel.RequestConvertible {
...
-                    .joinToString("&")
-        } ?: ""
-    }
-
-    companion object {
-        private fun defaultHeaders(): MutableMap<String, Any> = mutableMapOf("Accept-Encoding" to "compress;q=0.5, gzip;q=1.0")
-    }
+    private fun queryFromParameters(params: List<Pair<String, Any?>>?): String = params.orEmpty()
+            .filterNot { it.second == null }
+            .mapTo(mutableListOf()) { "${it.first}=${it.second}" }
```
kittinunf: Maybe using just `map` is better.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/FuelManager.kt
```diff
@@ -11,14 +11,9 @@ import java.security.KeyStore
 import java.util.concurrent.Executor
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.Executors
-import javax.net.ssl.HostnameVerifier
-import javax.net.ssl.HttpsURLConnection
-import javax.net.ssl.SSLContext
-import javax.net.ssl.SSLSocketFactory
-import javax.net.ssl.TrustManagerFactory
+import javax.net.ssl.*
```
kittinunf: I prefer to do individual import though. What do you think?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/UploadTaskRequest.kt
```diff
@@ -3,87 +3,90 @@ package com.github.kittinunf.fuel.core.requests
...
+private const val BUFFER_SIZE = 1024
+private val CRLF = "\r\n".toByteArray()
+
+private fun guessContentType(name: String): String = try {
+    URLConnection.guessContentTypeFromName(name) ?: "application/octet-stream"
+} catch (ex: NoClassDefFoundError) {
+    // The MimetypesFileTypeMap class doesn't exists on old Android devices.
+    "application/octet-stream"
+}
+
```
kittinunf: Can you please remove empty lines at the end?

kittinunf:
On the first glance, this looks pretty awesome! Thanks, @yoavst for doing this.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
yoavst: Have no idea why I used it like that. Fixing.

yoavst:
@kittinunf I've fixed #220 , tell me if you are OK with the fix or I'll revert and merge.

yoavst:
@kittinunf?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
```diff
@@ -1,47 +1,48 @@
...
         var modifiedPath = path
         var data: String? = null
-        val headerPairs: MutableMap<String, Any> = defaultHeaders()
-        if (encodeParameterInUrl(method)) {
-            var querySign = ""
-            val queryParamString = queryFromParameters(parameters)
-            if (queryParamString.isNotEmpty()) {
-                if (path.count() > 0) {
-                    querySign = if (path.last() == '?') "" else "?"
+        val headerPairs: MutableMap<String, Any> = defaultHeaders.toMutableMap()
```
kittinunf: Shall we drop the explicit type declaration?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
```diff
@@ -65,23 +66,14 @@ class Encoding : Fuel.RequestConvertible {
...
-    }
-
-    companion object {
-        private fun defaultHeaders(): MutableMap<String, Any> = mutableMapOf("Accept-Encoding" to "compress;q=0.5, gzip;q=1.0")
-    }
+    private fun queryFromParameters(params: List<Pair<String, Any?>>?): String = params.orEmpty()
+            .filterNot { it.second == null }
+            .joinToString("&") { "${it.first}=${it.second}" }
 
+    private val defaultHeaders: Map<String, Any> = mapOf("Accept-Encoding" to "compress;q=0.5, gzip;q=1.0")
```
kittinunf: Here as well.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
```diff
@@ -20,66 +20,64 @@ import java.util.concurrent.Future
...
+        var type: Type = Type.REQUEST,
+        val headers: MutableMap<String, String> = mutableMapOf(),
+        val parameters: List<Pair<String, Any?>> = listOf(),
+        var name: String = "",
+        val names: MutableList<String> = mutableListOf(),
+        val mediaTypes: MutableList<String> = mutableListOf(),
+        var timeoutInMillisecond: Int = 15000,
+        var timeoutReadInMillisecond: Int = timeoutInMillisecond) : Fuel.RequestConvertible {
+
+    @Deprecated(replaceWith = ReplaceWith("method"), message = "http naming is deprecated, use 'method' instead")
```
kittinunf: 👏

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -1,11 +1,7 @@
 package com.github.kittinunf.fuel.toolbox
 
 import com.github.kittinunf.fuel.Fuel
-import com.github.kittinunf.fuel.core.Client
-import com.github.kittinunf.fuel.core.FuelError
-import com.github.kittinunf.fuel.core.Method
-import com.github.kittinunf.fuel.core.Request
-import com.github.kittinunf.fuel.core.Response
+import com.github.kittinunf.fuel.core.*
```
kittinunf: I think we should perfer the import to be expanded.

kittinunf:
I have final check on this and commented with some small nits. After you fixed them, I think we could merge.


Thanks @yoavst for your hardwork!!! 👏 ✨

yoavst:
Back from the trip on the 31th, will do it then, using the new debugger duck I bought 😀

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/FuelError.kt
```diff
@@ -1,11 +1,6 @@
...
 
-import kotlin.properties.Delegates
-
-class FuelError : Exception() {
-    var exception: Exception by Delegates.notNull()
-    var errorData = ByteArray(0)
-    var response = Response()
-
+class FuelError(val exception: Exception, val errorData: ByteArray = ByteArray(0), val response: Response = Response.error())
+    : Exception("FuelError", exception) {
```
kittinunf: We should propagate the error so the info could be able to extract by clients. I suggest we use this version of the constructor `class FuelError(..) : Exception(exception)` instead.

build.gradle
```diff
@@ -21,7 +21,7 @@ allprojects {
         publish_version = '1.9.0'
 
         //dependencies version
-        kotlin_version = '1.1.3'
+        kotlin_version = '1.1.4'
```
kittinunf: Do you think we can just jump to version 1.1.4-2 altogether?

yoavst:
@kittinunf is it OK?

kittinunf:
👍 👍 👍 merge as soon as you are ready

PR #230, opened 2017-08-26
Modernize sample

kittinunf:
### What's in this PR?

This PR made a small improvement on the sample `MainActivity.kt`. Make it nicer and doesn't look like it gets abandoned.


kittinunf:
This is just a simple update. I will merge this in. :)

PR #244, opened 2017-09-16
Fuel's module fuel-andorid don't have Result's depend

shaomaicheng:
add compile 'com.github.kittinunf.result:result:1.2.0' in fuel-andorid's gradle

### DIFF
--- a/fuel-android/build.gradle
+++ b/fuel-android/build.gradle
@@ -60,6 +60,7 @@ dependencies {
 
     testCompile "junit:junit:$parent.ext.junit_version"
     testCompile "org.robolectric:robolectric:$parent.ext.robolectric_version"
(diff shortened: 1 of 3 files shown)


shaomaicheng:
why？some people can not build this project...

yoavst:
But `fuel-android` depends on `fuel` which depends on `result`

kittinunf:
I don't think that we need to explicitly add `Result` into the dependency list. Can you explain more @shaomaicheng ?

kittinunf:
I'm closing this

leus:
Using maven:
```
		<dependency>
			<groupId>com.github.kittinunf.fuel</groupId>
			<artifactId>fuel</artifactId>
			<version>1.8.0</version>
		</dependency>

```
I don't see any library or jar containing a `Result` class. Where is this class located?

Issue #249, opened 2017-09-28
POST Parameter encoding


mbreithu:
I'm trying to send some parameters to my API (Laravel project) using POST. The request works fine, but if one of the parameters contains a '+' (plus) character, the character is replaced with a ' ' (space) once the request reaches the server! 

After digging into the Encoding class, i found that when using POST, the content-type header is set to: "application/x-www-form-urlencoded" but the params themselves are not checked for special (reserved) characters. For urlencoded params the + sign will be interpreted as a space.

Solution would be to check params for reserved characters and replace them with their corresponding percent-encoding.

romanstingler:
I'm having the exactly same behaviour.

I hope it isn't intended to escape by myself

kittinunf:
Ah ... this is an oversight on my part. Would you be so kind to open PR for a patch? 🙏

mbreithu:
Will do!

PR #250, opened 2017-10-02
UrlEncode POST parameters

mbreithu:
Parameters of post request were not urlencoded before writing them into the body. Which led to incorrect results on server side because reserved characters were not in %-format!

Now parameter keys and values are encoded before joining them to parameter string.

fixes: #249

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
@@ -1,7 +1,11 @@
 package com.github.kittinunf.fuel.core
 
 import com.github.kittinunf.fuel.Fuel
-import java.net.*
+import java.net.MalformedURLException
+import java.net.URI
+import java.net.URISyntaxException
+import java.net.URL
+import java.net.URLEncoder
 
 class Encoding(val httpMethod: Method,
                val urlString: String,
@@ -70,7 +74,8 @@ class Encoding(val httpMethod: Method,
 
     private fun queryFromParameters(params: List<Pair<String, Any?>>?): String = params.orEmpty()
             .filterNot { it.second == null }
-            .joinToString("&") { "${URLEncoder.encode(it.first,"UTF-8")}=${URLEncoder.encode("${it.second}","UTF-8")}"}
+            .map { (key, value) ->  URLEncoder.encode(key, "UTF-8") to URLEncoder.encode("$value", "UTF-8") }
+            .joinToString("&") { (key, value) -> "$key=$value" }
 
     private val defaultHeaders = mapOf("Accept-Encoding" to "compress;q=0.5, gzip;q=1.0")
 }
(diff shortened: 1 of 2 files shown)


mbreithu:
@kittinunf Just saw that the travis-ci tests fail. But when i run the tests locally they all pass! Do you have any ideas what the difference between the two test runs could be?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
```diff
@@ -73,7 +70,7 @@ class Encoding(val httpMethod: Method,
 
     private fun queryFromParameters(params: List<Pair<String, Any?>>?): String = params.orEmpty()
             .filterNot { it.second == null }
-            .joinToString("&") { "${it.first}=${it.second}" }
```
kittinunf: How about we make the transformation more explicit? like

``` Kotlin
params.orEmpty()
  .filterNot ( it.second == null }
  .map { (key, value) ->  URLEncoder.encode(key, "UTF-8") to URLEncoder.encode(value, "UTF-8") }
  .joinToString("&") { (key, value) -> "$key=$value" }
```

fuel/src/test/kotlin/com/github/kittinunf/fuel/EncodingTest.kt
```diff
@@ -130,4 +131,42 @@ class EncodingTest : BaseTestCase() {
...
+            request.bodyCallback?.invoke(request, this, 0)
+        }.toByteArray()
+
+        val bodyString = String(body)
+        print(bodyString)
+        assertThat(bodyString, isEqualTo("param1=value1&param2=value2"))
+    }
+
+    @Test
+    fun testPostRequestEncodingWithReservedCharacters() {
```
kittinunf: 👏

kittinunf:
Simple nits and this PR is good to merge.

kittinunf:
I just rerun the tests and it passed. So nothing to be done from your side. 😉

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
```diff
@@ -70,7 +74,8 @@ class Encoding(val httpMethod: Method,
 
     private fun queryFromParameters(params: List<Pair<String, Any?>>?): String = params.orEmpty()
             .filterNot { it.second == null }
-            .joinToString("&") { "${URLEncoder.encode(it.first,"UTF-8")}=${URLEncoder.encode("${it.second}","UTF-8")}"}
+            .map { (key, value) ->  URLEncoder.encode(key, "UTF-8") to URLEncoder.encode("$value", "UTF-8") }
```
kittinunf: Just `value` maybe? 😄

mbreithu: param list is of type Pair<String,Any> and URLEncoder expects parameter to be of type String

kittinunf: Ah ... you are right. Thanks!

mset:
I'd benefit from this fix as well. Can we get it merged please?

yoavst:
@kittinunf ?

kittinunf:
Oh sure!

kittinunf:
@mbreithu, Thanks for your contribution

@mset, you can use the latest version of Fuel that includes the fix. 😉

Issue #116, opened 2016-10-08
JRE 7 required to use Basic Authentication


nyfd:
https://github.com/kittinunf/Fuel/releases/tag/1.3.1

Using Basic Authentication on Oracle JRE 6 (Debian) fails with an exception.

Apparently Basic Authentication uses Base64 encoding classes which are are compiled with target compatibility JVM 7.

See com.github.kittinunf.fuel.util.Base64.class

yoavst:
In a library of mine I use:
```
sourceCompatibility = 1.6
targetCompatibility = 1.6
```

But I don't know if it really works.

nyfd:
It should work, and it's easy to rebuild from sources with JRE compatibility set to 1.6

The problem is that the published JAR (https://github.com/kittinunf/Fuel/releases/tag/1.3.1) already contains those Base64-classes that require JRE 7 to run, and we prefer to use published artefacts from JCenter repository instead of building from sources.

nyfd:
Hmm, I'd like to understand how this could happen - I guess that the release was built on a system with JDK 7 installed as default Java.

Kotlin compiler produces class files targeted for JVM 6 automatically. 

Base64.java seems to be the only class written in pure Java so JDK 7 built it with target compatibility 7. 

You need to provide bootstrap path with JRE 6 runtime JARs when building for JRE 6 on a JDK 7/8/9 machine, which is not a big problem.
Probably it would be much easier to build the next release with JDK 6, if it's possible.

kittinunf:
I will do that for the next release.

kittinunf:
It is addressed by #123

kittinunf:
This should be fixed with latest version. 1.4.0

PR #253, opened 2017-10-09
Clean gradle build files

yoavst:
I moved the ext to buildscript, and moved kotlin-gradle-script definition to the main build.gradle.
Hope it will pass CI :)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@
 
 buildscript {
     ext {
-        publish_version = '1.11.0'
+        publishVersion = '1.11.0'
 
         //dependencies version
         kotlinVersion = '1.1.51'
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,27 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+    ext {
+        publish_version = '1.11.0'
+
+        //dependencies version
+        kotlin_version = '1.1.51'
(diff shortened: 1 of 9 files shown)


fuel-android/build.gradle
```diff
@@ -32,11 +29,6 @@ android {
         }
     }
 
-    compileOptions {
```
kittinunf: I think we have added this to accommodate this issue #116. I am all in to get rid of this but would it be a problem for our users?

build.gradle
```diff
@@ -1,34 +1,19 @@
...
-    }
-}
-
-allprojects {
     ext {
         publish_version = '1.11.0'
 
         //dependencies version
-        kotlin_version = '1.1.4-3'
+        kotlin_version = '1.1.51'
```
kittinunf: Since you are on top of this already, should we change naming convention on this a bit to camelCase, just to make this consistent across the board?

build.gradle
```diff
@@ -1,6 +1,27 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+    ext {
+        publish_version = '1.11.0'
```
kittinunf: Last 🐍 case to be 🚿 ?

kittinunf:
👍 small fix then i think we re good

kittinunf:
:+1: :shipit:

PR #256, opened 2017-10-11
Gradle updates

sschuberth:
Please have a look at the individual commit messages for the details.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,26 +17,6 @@ buildscript {
 }
 
 allprojects {
-    ext {
-        publish_version = '1.11.0'
-
-        //dependencies version
-        kotlin_version = '1.1.4-3'
-        result_version = '1.2.0'
-        rxjava_version = '2.1.3'
-        archLifecycleVersion = '1.0.0-alpha9'
-
-        //android dependencies version
-        compile_sdk_version = 26
-        build_tool_version = '26.0.1'
-
-        //test dependencies version
-        junit_version = '4.12'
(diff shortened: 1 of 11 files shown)


sschuberth:
Looks like this will conflict with #253 once merged.

build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
     dependencies {
         classpath 'com.android.tools.build:gradle:2.3.3'
         classpath 'com.dicedmelon.gradle:jacoco-android:0.1.1'
-
+        classpath 'com.github.ben-manes:gradle-versions-plugin:0.15.0'
```
yoavst: Explanation?

yoavst:
I can merge your changes to my pr.  @kittinunf do we want the versions on `gradle.properties` or in main `build.gradle`?

build.gradle
sschuberth: Please see the commit message for that commit.

kittinunf:
In my project, I don't put the version of dependencies in `gradle.properties` but I might have overlooked something.

I would love to hear whether there is a big win of doing this?

sschuberth:
IMO, putting project-wide (static) properties in `gradle.properties` is cleaner. Like the file name says, this simply is where they belong to.

kittinunf:
Let's revisit this later as it looks stale.

sschuberth:
No idea why you believe this to look stale from my side. I was replying to your question, and you did not respond any further to my argument.

PR #261, opened 2017-10-28
Pass the expected object type through

ColinHebert:
Ensure that the response we get back from `responseObject` matches the type we expect.

At the moment any call to `responseObject()` (doesn't matter if we use `reponseObject<String>()`) ends with a `Result<Any, FuelError>`.


kittinunf:
Ah you are right ... as I am not using Gson myself. I am not aware of this problem.

Would you please be so kind adding one test case for this? Thanks! 🙏

ColinHebert:
Done

kittinunf:
Thanks for helping 🍻

PR #268, opened 2017-11-11
Fix type issue with usage with Jackson in sync mode

### LINKED ISSUE #265: JSON deserialization behaves differently in blocking vs async mode
Using Fuel and Fuel-Jackson v1.11.0.
I have a service returning a JSON UserProfile object mapped to a Kotlin class:
`    data class UserProfile(var name:String)
`
If I use Fuel in in async mode, the response gets properly deserialised to a UserProfile object:
```
   url.http_get().responseObject<UserProfile >(){request, response, result ->
           logger.info(result.value.name)
```

Whereas if I use Fuel in blocking mode, the responseObject signature returns a triplet with "result" being typed as Result<Any, FuelError> instead of Result<UserProfile, FuelError>.

Besides, result.value contains a HashMap<String, Any> with the response fields, and not a UserProfile object.

I can't see any reason why the blocking mode would behave differently from the async mode and not deserialize the response as expected?

kittinunf:
### What's in this PR?

- Address issue with type not passing along properly. 

Close #265

### DIFF
--- a/fuel-gson/src/main/kotlin/com/github/kittinunf/fuel/gson/FuelGson.kt
+++ b/fuel-gson/src/main/kotlin/com/github/kittinunf/fuel/gson/FuelGson.kt
@@ -1,15 +1,18 @@
 package com.github.kittinunf.fuel.gson
 
-import com.github.kittinunf.fuel.core.*
+import com.github.kittinunf.fuel.core.FuelError
+import com.github.kittinunf.fuel.core.Handler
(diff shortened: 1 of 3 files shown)


kittinunf:
@yoavst Please help me review this, thanks!

kittinunf:
I guess this is okay to merge?

kittinunf:
I am merging this in.

Issue #279, opened 2017-11-28
Redirect not followed if location header is lowercase


denniss17:
I'm trying to access an URL by using Fuel. The server returns a 302, with the redirect URL in a `location` header (as opposed to the `Location` header). I think the check should be case insensitive.

I'll see if I can create a PR

PR #280, opened 2017-11-28
Also check for lowercase `location` header on redirect

denniss17:
Fixes https://github.com/kittinunf/Fuel/issues/279

When there is a redirect status, fuel will first detect the default `Location` header. If it is not present, it will try to use the `location` header.

I looked into creating a test for this, but I did not see a simple way to make https://httpbin.org/ return a redirect with a custom header.

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/RedirectionInterceptor.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/RedirectionInterceptor.kt
@@ -14,7 +14,7 @@ fun redirectResponseInterceptor(manager: FuelManager) =
                         response.statusCode == HttpsURLConnection.HTTP_MOVED_TEMP ||
                         response.statusCode == 307   // 307 TEMPORARY REDIRECT - https://httpstatuses.com/307
                         ) {
-                    val redirectedUrl? = response.headers.containsKey("Location") ? response.headers["Location"] : response.headers["location"]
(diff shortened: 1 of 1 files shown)


fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/RedirectionInterceptor.kt
```diff
@@ -14,7 +14,7 @@ fun redirectResponseInterceptor(manager: FuelManager) =
                         response.statusCode == HttpsURLConnection.HTTP_MOVED_TEMP ||
                         response.statusCode == 307   // 307 TEMPORARY REDIRECT - https://httpstatuses.com/307
                         ) {
-                    val redirectedUrl = response.headers["Location"]
+                    val redirectedUrl? = response.headers.containsKey("Location") ? response.headers["Location"] : response.headers["location"]
```
kittinunf: Do you think if we could use `inline fun String.toUpperCase(): String` or `lowercase` counterpart to simplify our logic a bit?

kindraywind: We're almost there! @denniss17 
Let's change a minor thing 👍

denniss17: No, except when we lowercase all keys in the headers map

kittinunf:
I think this is reasonable. Merging this in.

PR #283, opened 2017-12-07
Fix synchronous call with moshi responseObject() to return T instead of Any

rwhite226:


babedev:
This looks nice 😃

kittinunf:
Great for me ... @rwhite226 Would you mind adding the test to check the type?

kittinunf:
I kinda wanna merge this in. I will open another PR separately for adding test for this.

Issue #195, opened 2017-07-14
The console application does not terminate after a successful request


anatolystansler:
I developing console application on Kotlin. If I make a request from the main thread and request finishes (successful or not, doesn't matter) application won't close for a minute. Is that normal behavior? Is there a way to stop execution right after result handler finishes its execution?

Request:
```
Fuel.get("/info").responseString { _, _, result ->
        when (result) {
            is Result.Failure -> {
                println("Failure")
            }
            is Result.Success -> {
                println("Success")
            }
        }
    }
```

SimplySerenity:
Just noticed the same problem. I'm guessing it's a problem with the jvm waiting for the async thread officially close, because if you use the non-async version it terminates immediately.

yoavst:
`Fuel` opens a daemon thread that prevents the JVM from shutdown.
You can try this code to shutdown `Fuel`, though it should probably provide a better masochism for that:
```
FuelManager.executor.shutdownNow()
```

Is someone familiar with how OKHttp handles that?

vaurelios:
Same as #99

markGilchrist:
can we close this?

anatolystansler:
Yes, @yoavst solution works. Thank you!

PR #287, opened 2017-12-21
Mark created threads as daemon thread so they can't hang the program.

### LINKED ISSUE #99: Hangs on at exit
Hangs on at exit when using async mode,

``` Kotlin
Fuel.get("http://www.google.com/").responseString { request, response, result ->
        println(result.getAs<String>())
    }

FuelManager.instance.executor.shutdown() // this line fixes problem
```

nanodeath:
Resolves #99 and #195.


kittinunf:
Perfect!! Thanks so much!

Issue #240, opened 2017-09-08
Flexible toString method for Request


pltanton:
`toString` method of Request looks great and pretty useful in many cases now. It is useful to use it for logging before sending request, but often, when you use upload method you want to send some binary file to the remote REST service and when you do something like `log.info(request.toString())` it produses tons of non-ascii binary trash for logfile or console.

It would be really great if output of binary data would reduced automatically by libraray or by some additonal parameters or some representational class, which would incapsulate current toString logic.

It could be part of #225

markGilchrist:
can we close this?

PR #303, opened 2018-01-24
Add response.guessContentType() and improve .toString()

kindraywind:
## What's in this PR?
- Add `guessContentType(): String` to `Response`. It either checks `Content-type` in response header or   from `byteArrayInputStream`
- A bit improve version of response.toString()
From this:

```
<-- 200 (http://httpbin.org/image/png?key=value)
                                                        Response : OK
                                                        Length : 8090
                                                        Body : ��#�I��b6��Xv�8N:e�C�}����`�^L�:�~�Mw�I��W..... (and more 8kb of nonsense bytedata)
``` 

To this:

```
<-- 200 (http://httpbin.org/image/png?key=value)
                                                        Response : OK
                                                        Length : 8090
                                                        Body : (8090 bytes of image/png)
```

### What _isn't_ in this PR?
 - flexible .toString() for `Request` #240  (Still figuring better solution for this.)

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Response.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Response.kt
@@ -34,18 +34,9 @@ class Response(
         }
     }
 
-    override  fun toString(): String {
-        val contentType = guessContentType()
-
-        val dataString = if (contentType.isNotEmpty() &&
-                            (contentType.contains("image/") ||
-                            contentType.contains("application/octet-stream"))) {
-             "$contentLength bytes of ${guessContentType()}"
-        } else if (data.isNotEmpty()) {
-             String(data)
-        } else {
-            "(empty)"
-        }
+    override fun toString(): String {
+        val contentType = guessContentType(headers)
+        val dataString = processBody(contentType, data)
 
         return buildString {
             appendln("<-- $statusCode ($url)")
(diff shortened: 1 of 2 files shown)


fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Response.kt
```diff
@@ -33,17 +34,41 @@ class Response(
...
-    override fun toString(): String = buildString {
-        appendln("<-- $statusCode ($url)")
-        appendln("Response : $responseMessage")
-        appendln("Length : $contentLength")
-        appendln("Body : ${if (data.isNotEmpty()) String(data) else "(empty)"}")
-        appendln("Headers : (${headers.size})")
-        for ((key, value) in headers) {
-            appendln("$key : $value")
+    override  fun toString(): String {
+        var dataString = "(empty)"
```
kittinunf: In Kotlin, a local variable can be late initialized. For example

``` Kotlin

 val abc: String
    
if (isSet) {
  abc = "hey"
} else {
  abc = "yo"
}
```

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Response.kt
```diff
@@ -33,17 +34,41 @@ class Response(
...
     }
 
+    fun guessContentType(): String {
+        val contentTypeFromHeaders = headers["Content-Type"]?.first()
+        if (contentTypeFromHeaders is String) {
+            return contentTypeFromHeaders
+        }
+
+        val contentTypeFromStream = URLConnection.guessContentTypeFromStream(ByteArrayInputStream(data))
+        return if(contentTypeFromStream.isNullOrEmpty()) "(unknown)" else contentTypeFromStream
```
kittinunf: [nit] Space after `if`

PR #309, opened 2018-02-11
Add ability to set timeout in FuelManager to resolve #94

### LINKED ISSUE #91: request JSON encode
So Fuel can parse a JSON response... but can it encode an object to JSON for a request?

### LINKED ISSUE #94: Allow setting timeout on FuelManager
Atm, the timeouts are hardcoded [here](https://github.com/kittinunf/Fuel/blob/c6e3b748a706dbee3cb37813676e3f7cdc1518f8/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt#L29). It would be nice to be able to modify the timeout for all requests on the `FuelManager.instance` object, rather than having to manually configure the timeout for each request.

gahfy:
This allows to use timeout and readtimeout directly in `FuelManager`


kittinunf:
I think this looks okay to be merged unless anyone see differently.

kindraywind:
lgtm let's merge this @kittinunf

PR #311, opened 2018-02-12
add httpString()

dobashi:
Request.toString() method shows headers first and body second.
I wanna see the Request as HTTP protocol like this:

> GET https://www.example.com/path
> 
> Content-Type: text/plain
> Accept-Encoding : compress;q=0.5, gzip;q=1.0
>
> Request body string
>


kittinunf:
I like this addition. 👍from me.

dobashi:
Sorry I missed.
> Request.toString() method shows headers first and body second.

Request.toString() method shows body first and headers second.

kittinunf:
I dont see anything against this. Let's merge it.

Issue #246, opened 2017-09-19
async requests using coroutines


zjuhasz:
#168 asks about coroutine support but the issue was closed by poster after a recommendation for an extension function a user could implement in their own code. Coroutines are perfectly suited for an http request library and I was wondering if there is any consideration being given to adding coroutine support to Fuel.

markGilchrist:
@zjuhasz  This is now added can we close this issue?

Issue #168, opened 2017-05-20
coroutine supported?


walfud:
Will you support coroutine in future?

chrisly42:
Support can be easily added in the meanwhile by adding helper functions like this one:


```
import com.github.kittinunf.fuel.android.core.Json
import com.github.kittinunf.fuel.android.extension.responseJson
import com.github.kittinunf.fuel.core.FuelError
import com.github.kittinunf.fuel.core.Request
import com.github.kittinunf.fuel.core.Response
import com.github.kittinunf.result.Result
import kotlinx.coroutines.experimental.suspendCancellableCoroutine

suspend fun Request.asyncResponse(): Triple<Request, Response, Result<ByteArray, FuelError>> = suspendCancellableCoroutine { cont ->
    response { request, response, result ->
        cont.resume(Triple(request, response, result))
        cont.invokeOnCompletion {
            if (cont.isCancelled) {
                request.cancel()
            }
        }
    }
}
```

zjuhasz:
Why was this closed, was coroutine support added to Fuel?

PR #313, opened 2018-02-23
#246, #168: Add Kotlin Coroutines support

lucasvalenteds:
This pull requests adds some suspend functions to the Request class. They were added in a new module called `fuel-coroutines` and are intended to allows the developers to avoid callbacks while using asynchronous calls.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,6 +21,7 @@ buildscript {
         runnerVersion = '1.0.0'
         rulesVersion = '1.0.0'
         espressoVersion = '3.0.0'
+        kotlinCoroutinesVersion = '0.22.5'
         forgeVersion = '0.3.0'
     }
 
--- a/build.gradle
+++ b/build.gradle
@@ -9,6 +9,7 @@ buildscript {
         resultVersion = '1.3.0'
         rxjavaVersion = '2.1.8'
         archLifecycleVersion = "1.0.0"
+        kotlinCoroutinesVersion = '0.22.2'
 
         //android dependencies version
         fuelCompileSdkVersion = 26
@@ -21,7 +22,9 @@ buildscript {
         gsonVersion = '2.8.2'
         jacksonVersion = '2.9.3'
         moshiVersion = '1.5.0'
-        kotlinCoroutinesVersion = '0.22.2'
+        runnerVersion = '1.0.0'
+        rulesVersion = '1.0.0'
+        espressoVersion = '3.0.0'
     }
 
     repositories {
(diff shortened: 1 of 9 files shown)


yoavst:
Do we want to allow the user to select in which coroutine pool it will make the network request?

lucasvalenteds:
Do you mean wrapping the suspend function with `async { }` or `launch { }`? Something like:

```kotlin
fun main(args: Array<String>) = runBlocking {
    val call = async(CommonPool) { "https://httpbin.org/ip".httpGet().awaitStringResult() }
    val ipInJson = call.await()
    println(ipInJson)
}
```

iNoles:
Android has own coroutine context for UI too.

raharrison:
Doesn't Fuel already run in it's own ExecutorService?  We don't want to start the coroutine on another thread which in turn calls Fuel which passes the actual request onto another thread.

lucasvalenteds:
@iNoles I've added a Espresso test to covers the Android use case you've mention.

@raharrison The [`Request`](https://github.com/kittinunf/Fuel/blob/master/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt#L74) class uses a Future internally and also the `ExecutorService` you've mentioned. Since these fields are final and private I couldn't find a way to change them while implementing coroutines as a module apart from `core`.

kittinunf:
@lucasvalenteds First of all, thanks for this!! It has been a pretty long awaited feature.

What do you think about relaxing them so that it can be modified?

lucasvalenteds:
@kittinunf Your welcome. Thank you for the Fuel project!

About your question, I can't see how to make this coroutines module more flexible. I'm also failing to understand what it means in terms of implementation. Could you provide a use case to clarify how the developer can modify the module to get a better usage experience?

raharrison:
Can this get merged as it still presents a solution to using Fuel with coroutines, even if more enhancements are needed in the future?

kittinunf:
@lucasvalenteds would you be so generous applying the finishing touch?

lucasvalenteds:
@kittinunf Sure. Tell me what you need.

kittinunf:
it looks like `kotlinCoroutinesVersion` is missing from the `build.gradle` file?

lucasvalenteds:
You are right, @kittinunf. Sorry about that.

lucasvalenteds:
The CI has failing due to tests related to `fuel` module, which was not changed in this pull request. Do you have any idea why? I've run the `fuel-coroutines` tests and they are passing.

screenshot from 2018-04-28 06-36-59

kittinunf:
Let's merge this in. I will investigate from master.

PR #318, opened 2018-02-28
FuelRouting / Add support for Json request

matteocrippa:
- Added support to JSON request to FuelRouting allowing injecting value in `body`
- Updated readme
- Updated test


matteocrippa:
Is that something I can do to help in proceeding?

kittinunf:
Ah, it looks good to me. But could you please add some tests to make our codecov a bit more happier? Thanks!

matteocrippa:
I will close this and make a new one

PR #329, opened 2018-04-03
Add json request support in router

matteocrippa:
I added `body` access to allow this

### DIFF
--- a/fuel/build.gradle
+++ b/fuel/build.gradle
@@ -15,6 +15,7 @@ targetCompatibility = JavaVersion.VERSION_1_6
 dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
     compile "com.github.kittinunf.result:result:$resultVersion"
+    compile 'org.json:json:20170516'
 
     testCompile "junit:junit:$junitVersion"
 }
--- a/fuel/build.gradle
+++ b/fuel/build.gradle
@@ -15,8 +15,8 @@ targetCompatibility = JavaVersion.VERSION_1_6
 dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
     compile "com.github.kittinunf.result:result:$resultVersion"
-    compile 'org.json:json:20170516'
 
+    testCompile 'org.json:json:20170516'
     testCompile "junit:junit:$junitVersion"
 }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ language: android
 android:
   components:
     - tools
-    - build-tools-26.0.2
-    - android-26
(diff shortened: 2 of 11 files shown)


matteocrippa:
@kittinunf is it possible that tests for `RequestTest.kt` are failing?

Seems:

```
 timeoutInMillisecond = 15000,
                timeoutReadInMillisecond = 15000
```

were missing

kittinunf:
Yes, you are right. I am fixing the test in this PR #334. After that was merged, can you please rebase against new master?

matteocrippa:
@kittinunf ok, ping me if you notice that I'm missing the merge :)

kittinunf:
#334 It got merged. Can you rebase? And I will be happy to see this PR get merged 👍 🙏

fuel/build.gradle
```diff
@@ -15,6 +15,7 @@ targetCompatibility = JavaVersion.VERSION_1_6
 dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
     compile "com.github.kittinunf.result:result:$resultVersion"
+    compile 'org.json:json:20170516'
```
kittinunf: Hmm, do we use this somewhere?

matteocrippa: Yep, we need it for tests to manipulate and create a json, dunno why I was not able to get any other access to JSON

matteocrippa:
@kittinunf it seems it is now the Jackson test failing

```
com.github.kittinunf.fuel.FuelJacksonTest > jacksonTestResponseSyncObject FAILED
    com.github.kittinunf.fuel.core.FuelError at FuelJacksonTest.kt:190
        Caused by: com.github.kittinunf.fuel.core.HttpException at FuelJacksonTest.kt:190
com.github.kittinunf.fuel.FuelJacksonTest > jacksonTestResponseSyncObjectError FAILED
    java.lang.AssertionError at FuelJacksonTest.kt:119
```

fuel/build.gradle
kittinunf: For tests? Then, perhaps using `testCompile` is better?

matteocrippa: I need that to create a JSON object to test sending that as body

```
val json = JSONObject()
                         json.put("id", this.value)
                         json.toString()
```

kittinunf: Yes, here right?

```
  is GetParamsTest -> {
    val json = JSONObject()
    json.put("id", this.value)
    json.toString()
```

But it is only in the test right? So you can technically just do ` testCompile 'org.json:json:20170516'` so that it doesn't forcefully push dependencies to the users.

matteocrippa: Yep, fixed switching it to `testCompile`

PR #339, opened 2018-04-29
Apply final touch

kittinunf:
### What's in this PR?

- This PR applies final touch to the project so that gradle files look correct.
- Also, trigger test to make sure they are passing again, (we need to think about a way to make test less flaky)


kittinunf:
@lucasvalenteds The test should be working now ✅

PR #350, opened 2018-05-17
Correct read timeout instruction

babedev:
Correct the instruction that show how to set up read timeout.


kittinunf:
This is only README update right? 👍  Gogogo.

babedev:
Then I will merge this PR 😁

PR #354, opened 2018-05-27
An extra space was causing InterceptorTest test to always fail

ParkerK:
Noticed that this test was always failing. The test was expecting `"user-agent": "Fuel"`, but the response was returning  `"user-agent":"Fuel"`.

The missing space after the `:` was causing it to always return a failure.


kittinunf:
Wow thanks so much for this!!! ♥️♥️

ParkerK:
You're welcome, thanks for this library!

PR #367, opened 2018-06-23
Send Files at end of stream is better

TomasKit:


TomasKit:
Most of server apps don't care the order of upload fields, basically all business statements execute after upload stream complete.

BUT, In some scenes, say we are going to check a signature field at the beginning of the upload stream. We can not do that if the file field is the one.

There is an example: Alibaba Cloud's OSS upload service limit the file field should be the last one of all form fields.  

References: 
https://stackoverflow.com/questions/7449861/multipart-upload-form-is-order-guaranteed
https://www.alibabacloud.com/help/en/faq-detail/42976.html (The file must be the last field in those form fields. )

PR #371, opened 2018-07-02
added composite disposable to main activity 

markGilchrist:
I saw this and thought we should do this as a beginner will not understand why there is in error in AS and may not see how to resolve it


kittinunf:
I think it looks good to me, thanks! 👍

PR #374, opened 2018-07-05
Add result and object implementations

markGilchrist:
I had some ideas for error handling on coroutines, however there is an other pr by the original author covering the same changes.

The purpose of this pr is to show any parts that may be useful, if not the pr an branch will be junked

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ buildscript {
         publishVersion = '1.13.0'
 
         //dependencies version
-        kotlinVersion = '1.2.40'
+        kotlinVersion = '1.2.50'
         resultVersion = '1.4.0'
         rxjavaVersion = '2.1.12'
         archLifecycleVersion = "1.1.1"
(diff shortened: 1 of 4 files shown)


lucasvalenteds:
Thank you for working on this, @markGilchrist. I think having an API that returns a `Result` instance can be good addition too.

Although I'd like to share some thoughts before choosing the deprecation approach, as much as I regret the current naming of coroutines-based API.

Could you share your thoughts on the API described below? Using that approach the deprecation can be avoided. This PR could provide the `awaitResultString` function as well as the variants for `ByteArray`.

| Function | Return type |
| :--- | :--- |
| `awaitString()` | `Triple<Request, Response, Result<String, FuelError>>` |
| `awaitResultString()` | `Result<String, FuelError>` |
| `awaitStringResult()` | `String` |

lucasvalenteds:
Also, you've sad in Slack that you think documentation on the functions is useful, right? If that's the case, maybe some code snippets can be added in them. KDoc has tags to work with that.

What do you think?

PR #376, opened 2018-07-07
Add result and object implementations

markGilchrist:


fuel-coroutines/src/main/kotlin/com/github/kittinunf/fuel/coroutines/Coroutines.kt
```diff
@@ -1,13 +1,7 @@
-import com.github.kittinunf.fuel.core.Deserializable
-import com.github.kittinunf.fuel.core.FuelError
-import com.github.kittinunf.fuel.core.Request
+import com.github.kittinunf.fuel.core.*
```
kittinunf: [nits] Can we not collapse this? 🙏

kittinunf:
I think this looks awesome! I think if you can help me with one last comment, we can merge this. 🍺 

It looks like our test was not captured by Codecov though so our overall test coverage is reduced. But I wouldn't care about it too much as long as we can see them in the test.

Issue #360, opened 2018-06-03
how to pass use array parameter in a get request?


xinwo:
for example,
http://xx.xxx..x?a[]=1&a[]=2
usually means parameter a is an array, having 2 values - 1 and 2.

when i do below
val a = arrayOf(1, 2)
uri.httpGet(listOf("a" to a))

and check the log on server, it turns out something like below
a=%5BLjava.lang.String%3B%4095f9361

does that mean fuel doesn't support array parameter?

kittinunf:
Let's close this!

PR #378, opened 2018-07-08
Issue 360 add array support for http get request

markGilchrist:
This should fix issue #360 and mean we now support arrays in get requests


fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestTest.kt
```diff
@@ -7,7 +7,8 @@ import com.github.kittinunf.fuel.core.Method
 import com.github.kittinunf.fuel.core.Request
 import com.github.kittinunf.fuel.core.Response
 import org.hamcrest.CoreMatchers.*
-import org.junit.Assert.assertThat
+import org.json.JSONObject
+import org.junit.Assert.*
```
kittinunf: 🙏

markGilchrist: remove wild cards?

markGilchrist: Done

markGilchrist:
these the Jackson test are failing due to old kotlin version (I think) please can we look at my pr to update the version

fuel/src/main/kotlin/com/github/kittinunf/fuel/Fuel.kt
```diff
@@ -107,13 +124,15 @@ class Fuel {
...
-
     }
-
 }
 
 @JvmOverloads
-fun String.httpGet(parameters: List<Pair<String, Any?>>? = null): Request = Fuel.get(this, parameters)
+fun String.httpGet(parameters: List<Pair<String, Any?>>? = null): Request = Fuel.get(this, parameters?.flatMap { pair ->
+    (pair.second as? Array<*>)?.let {
+        it.map { "${pair.first}[]" to it }.toList()
```
kittinunf: Is this the standard way? I am happy to add into the lib if and only if it is a standard.

fuel/src/main/kotlin/com/github/kittinunf/fuel/Fuel.kt
```diff
@@ -107,13 +124,15 @@ class Fuel {
...
         fun request(convertible: RequestConvertible): Request = FuelManager.instance.request(convertible)
-
     }
-
 }
 
 @JvmOverloads
-fun String.httpGet(parameters: List<Pair<String, Any?>>? = null): Request = Fuel.get(this, parameters)
+fun String.httpGet(parameters: List<Pair<String, Any?>>? = null): Request = Fuel.get(this, parameters?.flatMap { pair ->
+    (pair.second as? Array<*>)?.let {
```
kittinunf: We wanna try to get a higher abstraction? Maybe `iterable`? So that `List<*>` can be used as well

markGilchrist: I will update

PR #381, opened 2018-07-09
Catch all deserializing errors

markGilchrist:
I have added a try/catch in deserializable to ensure that all errors are caught and returned as fuel errors


kittinunf:
Actually, `Result` has this capability built-in by using `Result.of`. @markGilchrist can you try that out instead?

markGilchrist:
@kittinunf  I have looked at that but Result.of has a fn signature of 
`fun <V : Any> of(f: () -> V): Result<V, Exception> `
and what I need is
`fun <V : Any> of(f: () -> V): Result<V, FuelError> `

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/TaskRequest.kt
```diff
@@ -19,5 +19,7 @@ internal open class TaskRequest(internal val request: Request) : Callable<Respon
             interruptCallback?.invoke(request)
         }
         throw error
+    } catch (exception:Exception){
+        throw  FuelError(exception)
```
kittinunf: [Nits] Extra space here.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/TaskRequest.kt
```diff
@@ -19,5 +19,7 @@ internal open class TaskRequest(internal val request: Request) : Callable<Respon
             interruptCallback?.invoke(request)
         }
         throw error
+    } catch (exception: Exception) {
+        throw  FuelError(exception)
```
kittinunf: [Nit] space bro 🙏

markGilchrist:
will merge in about 8 hours after work

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Deserializable.kt
```diff
@@ -74,9 +74,17 @@ private fun <T : Any, U : Deserializable<T>> Request.response(deserializable: U,
...
+fun <T : Any, U : Deserializable<T>> Request.response(deserializable: U): Triple<Request, Response, Result<T, FuelError>> {
+    var response : Response?  = null
+    val result = Result.of<T,FuelError> {
+        response = taskRequest.call()
+        deserializable.deserialize(response!!)
+    }
+    return Triple(this, response ?: result.component2()?.response!!, result)
 }
+
+fun <V : Any>Result.Companion.ofCatching(c : (exception:Exception) -> FuelError,f: () -> V): Result<V, FuelError> = try {
```
kittinunf: Hmm I don't see this being used anyware.

markGilchrist: I don't know why this is not showing as modified but it is

kittinunf: Can you tell me why the regular of is not suitable for this?

markGilchrist: Because when an error is thrown in deserialisation it will is likely to be as classCastExecption, which cannot be cast to a Fuel error, hence the exception is not throw. you need to pass in a lambda to describe how you change map all exception to the chosen error type,

I have raised a pr in Result, I will update it now with a failing test to demonstrate what I mean and you can then check out that branch?

more detail can be found here https://github.com/kittinunf/Result/pull/39

beylerian: We can't just replace 

> catch (error: FuelError)

with 
```kotlin
catch (ex: Exception)
``` 
https://github.com/kittinunf/Fuel/pull/381/files#diff-fececc99239d4e844c4cce8342aa7668L80

and forward it as 
```kotlin
Result.error(FuelError(error)))
```

?

beylerian: Maybe catch Exception in TaskRequest.call and wrap it as FuelError, and then catch it again in Request.response and wrap it again?

kittinunf:
I have raised #394 to improve on this

kittinunf:
Let's merge this to close the issue. 😉

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Deserializable.kt
```diff
@@ -69,14 +71,32 @@ private fun <T : Any, U : Deserializable<T>> Request.response(deserializable: U,
...
+                    it
+                } else {
+                    FuelError(it)
+                }
+            }
+
+    return Triple(this, response ?: Response.error(), result)
+}
+
+fun <V : Any> Result.Companion.ofCatching(c: (exception: Exception) -> FuelError, f: () -> V): Result<V, FuelError> = try {
```
markGilchrist: @kittinunf  shall we remove this before merging?

PR #385, opened 2018-07-12
updated gradle and kotlin version

markGilchrist:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Thu Jul 12 08:20:10 BST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.8-all.zip
```
kittinunf: Unfortunately, the publish plugin doesn't work well with Gradle 4.5 and above. 😢

markGilchrist: ok will revert

build.gradle
```diff
@@ -32,7 +32,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.3'
+        classpath 'com.android.tools.build:gradle:3.3.0-alpha02'
```
kittinunf: I think if we use 3.3.0, we need to update gradle version too? I guess we might limit this PR to just purely update Kotlin version.

PR #387, opened 2018-07-15
Add option to define coroutineContext for coroutine functions

markGilchrist:
this will prevent networkonmainthread exception


avaitla:
Was the CommonPool context in the coroutines code before and accidentally removed or is there some other reason it needs to be here for the caller to only invoke launch(UI)?

kittinunf:
I think this is reasonable to merge. Do you think we should add test for checking on the execution thread as well? So that we could have tests that cover us in the future.

markGilchrist:
@kittinunf  I have had a look at writing a test for the execution thread and I have run into a bit of a dead end. if you perform a test inside runBlocking the that forces the scope to be run on the blocking thread. 

If anyone has any wise ideas they would be very gratefully received

kittinunf:
I haven't had any good idea either ... let's merge it.

Issue #106, opened 2016-07-28
Progress not updating


silverboyir:
Hi
i want update progress of ProgressDialog in progress method in download
but progress not updating on view , im seeing the percent in logcat 

progress=new ProgressDialog(activity);
            progress.setMessage(activity.getResources().getString(R.string.download_video));
            progress.setProgressStyle(ProgressDialog.STYLE_HORIZONTAL);
            progress.setIndeterminate(true);
            progress.setProgress(0);
            progress.setMax(100);
            progress.setIndeterminate(false);
            progress.show();

and in progress

int currentProgress = (int)(readBytes \* 100.0 / totalBytes + 0.5);
                    Log.d(BadansazApplication.TAG, "download Pogress :"+String.valueOf(currentProgress));
                    progress.setProgress(currentProgress);

kittinunf:
Have you tried wrapped with `runOnUiThread`? View cannot update outside UI thread.

kittinunf:
This issue seems to be staled. Please re-open it if the issue persists.

PR #393, opened 2018-07-20
Added header support for multiple values for same key

### LINKED ISSUE #343: Can't add multiple headers with the same name (cookies)
The code:

```
val (request, response, result) = reqUrl.httpGet()
                        .header(Pair("Cookie", "val1=x"),
                                Pair("Cookie","val2=y"),
                                Pair("Cookie","val3=z"))
                        .responseString()
```

Expected functionality:

All cookies are included in request

Actual functionality:

Only the last Pair with val3 is included in the request

This happens because, in Request.kt line 28, the headers variable is a MutableMap, instead of perhaps an ArrayList of Pairs or a similar datatype.

Multiple headers with the same name is allowed in the [HTTP spec, section 4.2](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)

markGilchrist:
In this pr I hope I have a solution to the issue 342

close #343


fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestHeaderTest.kt
```diff
@@ -46,4 +47,22 @@ class RequestHeaderTest : BaseTestCase() {
         assertThat(string, containsString(headerValue))
     }
 
+    @Test
+    fun multipleHeadersByTheSameKeyWillBeCorrectlyFormatted(){
+        val request = manager.request(Method.GET, "/get")
+                .header("foo" to "bar","a" to "b", "cookie" to "val1=x", "cookie" to "val2=y","cookie" to "val3=z", "cookie" to "val4=j")
+
+        assertEquals("[ val1=x,val2=y,val3=z,val4=j ]",request.headers["cookie"] )
```
SleeplessByte: I think these brackets are just here for the `toString` and not actually when you send this request?, if so, that should be made clear in a comment or something like that :)

fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestHeaderTest.kt
```diff
@@ -46,4 +47,22 @@ class RequestHeaderTest : BaseTestCase() {
...
+        assertEquals("bar",request.headers["foo"] )
+        assertEquals("b",request.headers["a"] )
+    }
+
+    @Test
+    fun multipleHeadersByTheSameKeyWillShowLastUsingMap(){
+        val request = manager.request(Method.GET, "/get")
+                .header(mapOf("cookie" to "val1=x", "cookie" to "val2=y"),false)
```
SleeplessByte: Is this because `mapOf` does that? Probably should need an entry in the usage on `README.md`. Multiple values for a header? This is how you do it:

fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestHeaderTest.kt
markGilchrist: @SleeplessByte  thanks for the feed back I can remove the brackets and the code that adds them.

 to be clear these are NOT how the request should be formatted?

`cookie: [foo=bar,buzz=fizz]`
 
should be 

`cookie: foo=bar,buzz=fizz`

I can update this pr would you comment on other changes?

fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestHeaderTest.kt
markGilchrist: @SleeplessByte 
The problem with map is, (as I am sure you are aware), when you add multiple values for a key, they are all over written by the last one to write

SleeplessByte: Yep! So we/you should document that, as it's "logical" if you realise it :)

fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestHeaderTest.kt
SleeplessByte: Hi @markGilchrist , No, you got it! I commented on the complete PR. Rest looked great.

I would add a space. It's not wrong to not add a space, but the format allows for it, so it's more readable: i.e. this would be `Cookie: foo=bar; buzz=fizz`. Note that I used a semi-colon, which is the correct delimiter. Here is the RFC spec:

https://tools.ietf.org/html/rfc7540#section-8.1.2.5

```
8.1.2.5.  Compressing the Cookie Header Field

   The Cookie header field [COOKIE] uses a semi-colon (";") to delimit
   cookie-pairs (or "crumbs").  This header field doesn't follow the
   list construction rules in HTTP (see [RFC7230], Section 3.2.2), which
   prevents cookie-pairs from being separated into different name-value
   pairs.  This can significantly reduce compression efficiency as
   individual cookie-pairs are updated.

   To allow for better compression efficiency, the Cookie header field
   MAY be split into separate header fields, each with one or more
   cookie-pairs.  If there are multiple Cookie header fields after
   decompression, these MUST be concatenated into a single octet string
   using the two-octet delimiter of 0x3B, 0x20 (the ASCII string "; ")
   before being passed into a non-HTTP/2 context, such as an HTTP/1.1
   connection, or a generic HTTP server application.

   Therefore, the following two lists of Cookie header fields are
   semantically equivalent.

     cookie: a=b; c=d; e=f

     cookie: a=b
     cookie: c=d
     cookie: e=f
```

kittinunf:
This looks good to me. What is wrong with the travis though? Let me rerun.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
```diff
@@ -90,10 +92,26 @@ class Request(
...
+     *
+     * */
     fun header(vararg pairs: Pair<String, Any>?): Request {
-        pairs.forEach {
-            if (it != null)
+        pairs.filterNotNull().forEach {
+            if (!headers.containsKey(it.first)) {
                 headers += Pair(it.first, it.second.toString())
+            } else {
+                headers[it.first] = headers[it.first]!!.let { value ->
```
kittinunf: ah ... can we not use `!!` though?

SleeplessByte: headers[it.first] returns a nullable, and since line #106 checks for that, this will never fail...
...except that it is absolutely not thread safe.

kittinunf: I would say that `fun <K, V> Map<K, V>.getValue(key: K): V` is probably better if we dont want to deal with nullable value.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
```diff
@@ -90,10 +92,26 @@ class Request(
...
+     *  will be sent in the format of `key` : [ value,value,value ] </p>
+     * @param pairs This all the key value pair you wish to add to the headers
+     *
+     * @return the request supplied
+     *
+     * */
     fun header(vararg pairs: Pair<String, Any>?): Request {
-        pairs.forEach {
-            if (it != null)
+        pairs.filterNotNull().forEach {
```
kittinunf: You might wanna do `forEach { (key, value) ->` so you don't have to keep calling `.first` or `.second`

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
SleeplessByte: Agreed!

fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestHeaderTest.kt
markGilchrist: @SleeplessByte how would you suggest we document this?

SleeplessByte: **Adding multiple values to a header**
If you want to mimic multiple values of a header such as:
```
Cookie: foo=bar
Cookie: baz=42
```

You can use:
```Kotlin
.header("Cookie" to "foo=bar", "Cookie" to "baz=42")
// => Cookie: foo=bar; baz=42
```

It is important to directly call `.header` and that you don't wrap the arguments with `mapOf`, because `mapOf` will simply overwrite any previously set value:

```Kotlin
.header(mapOf("Cookie" to "foo=bar", "Cookie" to "baz=42"))
// => Cookie: baz=42

File: fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt [hunk 5]
```diff
@@ -90,10 +92,26 @@ class Request(
...
-            if (it != null)
-                headers += Pair(it.first, it.second.toString())
+        pairs.filterNotNull().forEach { (key,value) ->
+            if (!headers.containsKey(key)) {
+                headers += Pair(key, value.toString())
+            } else {
+                headers[key] =  headers.getValue(key).let {
+                    var str = it
+                    if (str[0] != '[') str = "[ $str ]"
+                     str.replace(" ]", "; $value ]")
```
kittinunf: What is our stance on the `[` and `]`? According to the RFC, we don't need them?

markGilchrist: @kittinunf @SleeplessByte  I am happy to do whatever you feel is best, I have no strong feelings on this

fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestHeaderTest.kt
markGilchrist: I will add this to the readme and shall I add kdocs?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
SleeplessByte: They don't work! If you send a header with `[]` it will literally produce this:

```
Cookie: [val1=a; val2=b; val3=c]
```

output the following cookies pairs:
- `[val1=a`
- `val2=b`
- `val3=c]`

fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestHeaderTest.kt
SleeplessByte: That would be nice! I traverse source code often myself 🥉

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
markGilchrist: @SleeplessByte In my latest commit I have added a test to hit http and analyse what we get back. in you comment above do I need to remove the `;` and the `[]`?

fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestHeaderTest.kt
```diff
@@ -46,4 +47,48 @@ class RequestHeaderTest : BaseTestCase() {
         assertThat(string, containsString(headerValue))
     }
 
+    @Test
+    fun httpRequestHeaderWithDuplicateValuesForCookieKey() {
+        manager.request(Method.GET, "/get")
+                .header("foo" to "bar","a" to "b", "cookie" to "val1=x", "cookie" to "val2=y","cookie" to "val3=z", "cookie" to "val4=j")
+                .responseObject (HttpBinHeadersDeserializer()).third.fold({
+            assertEquals("[ val1=x; val2=y; val3=z; val4=j ]",it.headers["Cookie"])
```
markGilchrist: @SleeplessByte please let me know what the expected should be here?

SleeplessByte:
Oh sorry! I didn't get notifications for the last requests @kittinunf and @markGilchrist : 

I'm looking at what was merged and it looks good! No brackets and the delimiter is a semi-colon. Seems to me that this is as it should be 💯

markGilchrist:
@SleeplessByte Glad it is ok I am keen to look other issues now

PR #394, opened 2018-07-23
Improvement on the issue proposal

kittinunf:
### What's in this PR?

@markGilchrist This is my proposal to tackle the issue, can you please take a look and see whether this is something looks okay to accept?

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Deserializable.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Deserializable.kt
@@ -2,6 +2,8 @@ package com.github.kittinunf.fuel.core
 
 import com.github.kittinunf.fuel.core.requests.AsyncTaskRequest
 import com.github.kittinunf.result.Result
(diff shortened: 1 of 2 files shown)


fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Deserializable.kt
```diff
@@ -75,16 +77,22 @@ private fun <T : Any, U : Deserializable<T>> Request.response(deserializable: U,
...
+            .mapError {
+                (it as? FuelError)?.let { response = it.response }
+                FuelError(it)
+            }
+
+    return Triple(this, response ?: Response.error(), result)
 }
 
-fun <V : Any>Result.Companion.ofCatching(c : (exception:Exception) -> FuelError,f: () -> V): Result<V, FuelError> = try {
+fun <V : Any> Result.Companion.ofCatching(c: (exception: Exception) -> FuelError, f: () -> V): Result<V, FuelError> = try {
```
kittinunf: Maybe we don't really need this.

PR #401, opened 2018-07-31
Improvement to unit tests for download

markGilchrist:


kittinunf:
🤔 But the test is green though ...

screen shot 2018-07-31 at 23 26 57

PR #405, opened 2018-08-02
Issue 102 added test to verify behaviour is correct

markGilchrist:
written test that currently passes, I feel that appropiate behaviour is already implmented


kittinunf:
🤔 Which behavior do you try to write the test for though?

markGilchrist:
@kittinunf The tests recreates the behaviour raised in the issue, hence we get a  junction in the roads, as I see it

if the current behaviour is what we want, (handled as a fuel exception, which I think is correct) then we can either junk this pr as superfluous 

or we decided this is not the desired behaviour and we change the test and implement changes.

as ever I am here to do whatever you ask!

kittinunf:
Ah I see. Understood now. Sorry just need a bit more info. Approved!

kittinunf:
I will merge on your behalf. :)

PR #415, opened 2018-08-21
Added ByteArray property to handle binary data with FuelRouting

yhanada:
In this PR, I added byte array properties to FuelRouting to directly set ByteArray to message body when POSTing binary data such as Base64 and Protobuf data.

FuelRouting class will have ByteArray and String properties as message body, ByteArray takes precedence.

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/util/FuelRouting.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/util/FuelRouting.kt
@@ -27,6 +27,10 @@ interface FuelRouting: Fuel.RequestConvertible {
      */
     val params: List<Pair<String, Any?>>?
     /**
+     * Body to handle binary type of request (e.g. application/octet-stream )
+     */
+    val bytes: ByteArray?
+    /**
      * Body to handle other type of request (e.g. application/json )
      */
     val body: String?
@@ -49,7 +53,9 @@ interface FuelRouting: Fuel.RequestConvertible {
                     urlString = path,
                     parameters = params
             )
-            body?.let {
+            bytes?.let {
+                encoder.request.body(it)
+            } ?: body?.let {
                 encoder.request.body(it)
             }
             // return the generated encoder with custom header injected
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/util/FuelRouting.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/util/FuelRouting.kt
@@ -53,9 +53,9 @@ interface FuelRouting: Fuel.RequestConvertible {
                     urlString = path,
                     parameters = params
             )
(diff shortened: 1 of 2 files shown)


fuel/src/main/kotlin/com/github/kittinunf/fuel/util/FuelRouting.kt
```diff
@@ -26,6 +26,10 @@ interface FuelRouting: Fuel.RequestConvertible {
      * It uses a pair with String, Any.
      */
     val params: List<Pair<String, Any?>>?
+    /**
+     * Body to handle binary type of request (e.g. application/octet-stream )
+     */
+    val bytes: ByteArray?
```
kittinunf: I guess this looks like a better way of sending the binary data. Do you think we should deprecate the `body` in the string version and prefer this instead @yhanada ?

kittinunf:
Welcome to Fuel project and thanks for your first contribution!!! 🎉 🎉 🎉

fuel/src/main/kotlin/com/github/kittinunf/fuel/util/FuelRouting.kt
yhanada: thanks for your comment @kittinunf 
Do you think that Fuel is used string format such as JSON more often to POST?
If so, it may be early to deprelate string version.
Should I fix it as follows?

```
    body?.let {
        encoder.request.body(it)
    } ?: bytes?.let {
        encoder.request.body(it)
    }
```

kittinunf: Okay.

yhanada: Thanks I fixed it. 🙇

fuel/src/test/kotlin/com/github/kittinunf/fuel/RoutingTest.kt
```diff
@@ -163,4 +178,37 @@ class RoutingTest: BaseTestCase() {
...
+        assertThat(response, notNullValue())
+        assertThat(error, nullValue())
+        assertThat(data, notNullValue())
+
+        val statusCode = HttpURLConnection.HTTP_OK
+        assertThat(response?.statusCode, isEqualTo(statusCode))
+
+        val json = JSONObject(string)
+        val bytes = Base64.decode(json.optString("data"), Base64.DEFAULT)
+        assertThat(String(bytes), isEqualTo(paramValue))
```
kittinunf: Lastly, could you please test this against the response body coming from the server not only just the paramValue. I think httpbin return data back to us as well.

kittinunf:
I think it looks good, can you check on my last comment. If that one is fixed, I think we can merge this.

PR #417, opened 2018-08-22
add payload to delete and get

markGilchrist:


markGilchrist:
Line 456, of RequestTest.kt adds both query params and body, please let me know if this is not a suitable test

fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestTest.kt
```diff
@@ -735,10 +741,10 @@ class RequestTest : BaseTestCase() {
         request.responseString().third.fold({ string ->
             try {
                 val json = JSONObject(string)
-                assertEquals(json.getJSONObject("args").getString("bar"),lionel)
+                assertEquals(json.getJSONObject("args").getString("bar"), lionel)
                 assertEquals(json.getJSONObject("args").getJSONArray("foo[]").map { it.toString() },
-                list.toList())
-            } catch (e: Exception){
+                        list.toList())
```
kittinunf: [Super nits] Can we combine this with the previous line?

kittinunf:
Ah, I got it. Thanks. I left one small note, otherwise it is good.

fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestTest.kt
markGilchrist: done

PR #418, opened 2018-08-28
Add add json data method

markGilchrist:


kittinunf:
Could you rebase against the updated master? 🙏

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
```diff
@@ -145,6 +145,11 @@ class Request(
 
     fun body(body: String, charset: Charset = Charsets.UTF_8): Request = body(body.toByteArray(charset))
 
+    fun jsonBody(body: String, charset: Charset = Charsets.UTF_8): Request {
+        header( "content-type" to "application/json")
+        return body(body,charset)
```
kittinunf: [Nits] Space between `,` and `charset`

markGilchrist: your eye for detail is amazing, will change

Issue #134, opened 2017-03-01
Support Javascript backend


cah-andrewfitzgerald:
With the release of Kotlin 1.1, the Javascript backend is no longer experimental.
It would be nice if the same HTTP API was available at all layers of the kotlin stack.

kittinunf:
This is really awesome idea. It is also aligned to this library's vision. However, as discussed, I am no where near expert in JS backend and not sure how to make this happen. You are very happy to help contribute in here. 👍

yoavst:
I've started working on it, but not promise anything (taking a vacation trip next week, and start learning in university afterwards)

kittinunf:
@yoavst Awesome man! I am looking forward to seeing that pans out. :)

yoavst:
The branch will be https://github.com/kittinunf/Fuel/tree/multiplatform_feature

yoavst:
Report:
Currently kotlin multi platform feature is too buggy to convert an existing library.

In addition, `Fuel` is too connected with Java. Instead of working on the multi platform feature, I'll work a bit on hiding implementation details under interfaces.

Is it OK, @kittinunf ?

kittinunf:
OK 👍

joelhandwell:
From Kotlin/Native 0.6 it supports common code with ```expect``` and ```actual``` keywords. Can the scope of this issue be jvm / jvm-android / js / native-ios / native-macos / native-windows / native-windows, etc?

iNoles:
It is getting impossible to do this without removing a lot of code. It also Forced Coroutines to be a part of the core. That may be a huge turn off to some people.

yoavst:
@iNoles What's about trying to use the [official http client](https://ktor.io/clients/index.html) as backend, wrapping it with fuel?

iNoles:
Major Code Rewrite. Ktor is very heavily on Coroutines too.

SleeplessByte:
Do we really need a major code rewrite? It seems we can write a `Client` that uses the ktor client,

iNoles:
Okay, Fuel 3.x supports Javascript!

iNoles:
@yoavst Fuel 3.x SNAPSHOTS supports Javascript

bashor:
@iNoles do you compare it to ktor in terms of size?

PR #423, opened 2018-09-06
[WIP] Migrate Gradle scripts to Kotlin DSL

lucasvalenteds:
This pull requests rewrites all Gradle build files from Groovy DSL to Kotlin DSL.

The motivation to change them is to reduce duplication and maintenance effort by declaring versions in a single file and defining common logic in one place, which make them reusable. Better syntax highlight and autocomplete is expected too.

At the moment, all tests are passing and Android Studio doesn't have any warnings. 

**Room for improvement**
* Logic to build Android-based modules are not as reusable as JVM-based modules
* Make sure publish plugin is working as expected. At the moment Gradle is loggin the warning below.
  ```
  Publication release not found in project :fuel-android.
  Publication release not found in project :fuel-livedata
  ```
* Gradle found different versions of `kotlin-stdlib`, but tests are passing. The command below may help to identify them.
  ```
  ./gradlew dependencyInsight --configure compile --dependency kotlin-stdlib
  ```

I'd love hear some feedback from you about using Kotlin instead of Groovy and also suggestions to improve the build scripts!


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Fri Jul 27 18:14:34 BST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10-all.zip
```
markGilchrist: I think @kittinunf had a reason why we couldn't do this (sorry I can't remember what it was)

versions.gradle.kts
```diff
@@ -0,0 +1,28 @@
+buildscript {
+    // Library version
+    extra["publishVersion"] = "1.15.0"
```
markGilchrist: @lucasvalenteds sorry if this is a dumb question, but is this where you are setting all the values referenced in all the other build scripts?

markGilchrist:
@lucasvalenteds I really like this as an Idea, I am new to using Kotlin instead of groovy, could you add a link to the notes on this in your pr?

Do you know if jet brains / Google have any long term ambitions to move from groovy to Kotlin for build scripts?

iNoles:
@markGilchrist It was Gradle has long terms for it. 

It is in https://github.com/gradle/kotlin-dsl

gradle/wrapper/gradle-wrapper.properties
lucasvalenteds: Good point. Maybe it's an issue with the [novoda/bintray-release](https://github.com/novoda/bintray-release) plugin. The README file says it's not compatible with Gradle 4.5 at the moment.

iNoles: @lucasvalenteds Actually, it mentioned about "Currently Gradle 4.5 doesn't work with Android projects." It does have own Kotlin DSL examples too.

versions.gradle.kts
lucasvalenteds: Yes, the `version.gradle.kts` file contains all variables that was defined in `build.gradle` before the rewrite.

The `extra` variable is equivalent to the `ext` property in Groovy DSL. It's a container to store data that can be shared across multiple files.

To share them, the function `apply(from = "version.gradle.kts")` file is called in the root build script (`build.gradle.kts`) and propagated to the subprojects using the `subprojects {}` block.

lucasvalenteds:
Thank you for sharing your thoughts, @markGilchrist.

> could you add a link to the notes on this in your pr?

The source code is available at the link @iNoles mentioned. Besides that more information can be found at [jnizet/gradle-kotlin-dsl-migration-guide](https://github.com/jnizet/gradle-kotlin-dsl-migration-guide) which compares scripts in both DSLs to help migrations. There is also the #gradle channel in the Kotlin Slack and a KotlinConf 2017 [presentation](https://youtu.be/lg5hPBEIQ2E?t=27m34s) by a core developer in the Gradle Kotlin DSL project.

> Do you know if jet brains / Google have any long term ambitions to move from groovy to Kotlin for build scripts?

Sorry, I can't remember any statements about that.

gradle/wrapper/gradle-wrapper.properties
lucasvalenteds: Thank you for the clarification, @iNoles. I'll look into a workaround.

kittinunf: Yes, it has to be 4.4 at the moment or we get rid of `bintray-release`. But I would love to do that in the different PR.

kittinunf:
First of all, thanks so much for taking a stab on this. I really like the idea though. 

Just one suggestion though, instead of having version.gradle.kts for our dependency, what do you think about using `buildSrc` for this, I believe it should simplify the build script even further.

To elaborate, you can add

1. create `buildSrc` folder
2. create `build.gradle.kts` file with this content

``` Kotlin
plugins {
    `kotlin-dsl`
}
```
3. create `src/main/java` with a file that control our dependencies version

``` Kotlin
object Versions {
  const val foo = "1.2.3"
  const val bar = "2.3.4"
}
```
[Optional]
4. It might be worthwhile to also include the string dependencies too like

```
object Deps {
  val rxJava = "io.reactivex.rxjava2:rxjava:${Versions.rx}"
}
```

build.gradle.kts
```diff
@@ -0,0 +1,91 @@
...
+
+        tasks.withType<JacocoReport> {
+            reports {
+                html.isEnabled = false
+                xml.isEnabled = true
+                csv.isEnabled = false
+            }
+        }
+    }
+
```
kittinunf: [Q] Do you think we can just move Android module configuration in here too? it looks a bit more consistent (maybe) to have them in one place?

```
subproject {
  //..
  
  if (isAndroid) {
    //configure build configuration here.
  }
}
```

build.gradle.kts
```diff
@@ -0,0 +1,91 @@
...
+    repositories {
+        mavenCentral()
+        jcenter()
+        google()
+    }
+}
+
+subprojects {
+    val isAndroidModule = project.name == "fuel-android" || project.name == "fuel-livedata"
+    val isSample = project.name.contains("sample")
```
kittinunf: Shall we add `"sample-java"` too?

build.gradle.kts
```diff
@@ -0,0 +1,91 @@
...
+
+    repositories {
+        mavenCentral()
+        jcenter()
+        google()
+    }
+}
+
+subprojects {
+    val isAndroidModule = project.name == "fuel-android" || project.name == "fuel-livedata"
```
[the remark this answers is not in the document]
lucasvalenteds: Changed in 9cb8e7a

build.gradle.kts
lucasvalenteds: I think it's OK for modules that are libraries (`com.android.library` plugin). The commit e2fced7 reuses `fuel-android` and `fuel-livedata` as you've proposed.

build.gradle.kts
lucasvalenteds: Sorry, I'm not sure what you mean. The `isSample` variable is used to not apply the Bintray release plugin in the samples modules. Also now it's specified in the change 9cb8e7a.

lucasvalenteds:
Thank you for taking the time to review this, @kittinunf!

Also thanks for pointing the `buildSrc` mechanism, I didn't know about that. The commit 4455518 was changed according to your suggestion.

> do you know why our codecov got messed up heavily though?

I'm not sure, but it seems to be the merge commit (322421b).

lucasvalenteds:
Also, I would like to know what version of the JDK/Android the library should target. It's not clear to me if the compatibility should be set to Java 6 or Java 8.

iNoles:
@lucasvalenteds I noticed about Base64.java's coverage is 44.18% <0%> (-27.71%)

build.gradle.kts
```diff
@@ -0,0 +1,140 @@
...
+
+buildscript {
+    repositories {
+        mavenCentral()
+        jcenter()
+        google()
+    }
+
+    dependencies {
+        classpath("com.android.tools.build:gradle:3.1.3")
```
kittinunf: Can we also move this into `dependencies.kt` too? maybe we can call `BulidScript.androidGradle`? Any specific reason you want it to hang around here? 😄

fuel-moshi/build.gradle.kts
```diff
@@ -0,0 +1,5 @@
+plugins { java }
+dependencies {
```
kittinunf: [Nits] Can you add extra new line between `plugins` and `dependencies` like counterparts?

fuel/build.gradle.kts
```diff
@@ -0,0 +1,10 @@
+import org.jetbrains.kotlin.gradle.dsl.Coroutines
+import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
+import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
```
kittinunf: [Q] I don't think these imports are necessary? 🤔

kittinunf:
This is nice 🎉 ... but at the end, unfortunately, we need to change the gradle version back to 4.4 at the moment so that it won't affect the publish step. 😢

lucasvalenteds:
@kittinunf I understand that changing the release tooling may not be appropriate at this time, not just because of the PR scope but also the release process itself.

As a side note, sticking with Gradle 4.4 blocks the issue #134. I've successfully migrated the Result library to multiplatform (JVM, JS and native), but a newer Gradle version is required in that project too.

Are you open to a new PR to change de publish plugin? I can try to find another plugin later and close this issue in case there's no other option. :+1:

fuel/build.gradle.kts
lucasvalenteds: Good catch. They're not necessary now, but they were in the first commits. Removed in db7772e.

fuel-moshi/build.gradle.kts
lucasvalenteds: Changed in b6392d8.

build.gradle.kts
lucasvalenteds: Yes, we can. The commit 76388d3 moves them to `Plugins` object. The name can be changed if you think `BulidScript` is better. :+1:

kittinunf:
Ah, my suggestion is to use 4.4 for now? Then, we have another PR upgrading to the newer version to 4.9 and also fixing the publishing plugin at the same time? If this does make sense for you, let me know.

> I can try to find another plugin later

I tried 3 or 4 of them and find the [novoda/release](https://github.com/novoda/bintray-release) to be the sanest one to be used, the rest of them the configuration parts are either so long or so cryptic.

kittinunf:
So, let's change this back to 4.4 for now and merge? I will find the suitable solution for us to move the better/newer gradle version.

lucasvalenteds:
@kittinunf Sorry for the delay. The commit ce7dcbcbe5156addb6a2d85befd04bbd9a427959 reverts the Gradle version to 4.4.

> another PR upgrading to the newer version to 4.9 and also fixing the publishing plugin at the same time? If this does make sense for you, let me know

:+1:

PR #424, opened 2018-09-07
Integration with Project Reactor

lucasvalenteds:
This PR adds a new module that allows users to use [Mono](https://projectreactor.io/docs/core/milestone/reference/docs/index.html#mono) class from [Project Reactor](https://github.com/reactor/reactor-core) in projects written using reactive programming style targeting JVM.

It follows the same conventions used throughout the project for naming and error handling. There are functions `monoOfResponse`, `monoOf<Type>` and `monoOfResult<Type>` as in RxJava and Coroutines modules.

Please, let me know if you have any comments!


fuel-reactor/build.gradle
```diff
@@ -0,0 +1,43 @@
+apply plugin: 'kotlin'
```
kittinunf: I think this needs to be modified a bit after the Kotlin-dsl made it into master #423, right?

iNoles:
```
plugins {
   kotlin("jvm") 
}
```

lucasvalenteds: Changed in e1180d2d05b6ad2dc8e2ed0c8fcc6165fc57a7f2

fuel-reactor/src/main/kotlin/com/github/kittinunf/fuel/reactor/Reactor.kt
```diff
@@ -0,0 +1,51 @@
...
+fun Request.monoOfBytes(): Mono<ByteArray> =
+    monoOfResultFold(ByteArrayDeserializer())
+
+fun Request.monoOfString(charset: Charset = Charsets.UTF_8): Mono<String> =
+    monoOfResultFold(StringDeserializer(charset))
+
+fun <T : Any> Request.monoOfObject(mapper: Deserializable<T>): Mono<T> =
+    monoOfResultFold(mapper)
+
+private fun <T : Any> Request.monoOfResultUnFolded(mapper: Deserializable<T>): Mono<Result<T, FuelError>> =
```
kittinunf: Can this method and `monoOfResultFold` use one private method? Practically, `monoOfResultFold` is a map version of  `monoOfResultUnFolded`?

fuel-reactor/src/main/kotlin/com/github/kittinunf/fuel/reactor/Reactor.kt
```diff
@@ -0,0 +1,51 @@
...
+import java.nio.charset.Charset
+
+private fun <T : Any> Request.monoOfResultFold(mapper: Deserializable<T>): Mono<T> =
+    Mono.create<T> { sink ->
+        sink.onCancel(this::cancel)
+        val (_, _, result) = response(mapper)
+        result.fold(sink::success, sink::error)
+    }
+
+fun Request.monoOfBytes(): Mono<ByteArray> =
```
kittinunf: [Q] Shall we omit the `Of` here so it kinda goes similar to the `responseXXX` version? 

`responseBytes` => `monoBytes`
`responseString` => `monoString`
`responseObject` => `monoObject`

lucasvalenteds: I prefer `monoOf`, but I agree that using `mono` only can be better for consistency. Changed in f669d3c.

fuel-reactor/src/main/kotlin/com/github/kittinunf/fuel/reactor/Reactor.kt
lucasvalenteds: Changed in e9add41.

fuel-reactor/src/main/kotlin/com/github/kittinunf/fuel/reactor/Reactor.kt
SleeplessByte: `monoOf<T>` would still work, and agreed with all the other comments 🗡

kittinunf:
I think it looks great to me. Regarding the test, we might need @SleeplessByte to shed some shiny light ✨ ⚡️ to make it inline with the mock server thingy though. 

Also, it could be done in the follow-up PR.

iNoles:
ReadmeSample**.kt should not be needed for this project.

SleeplessByte:
@kittinunf if you take a look at #444, it should be clear how to use mock-server for this. In general:

1. add the depedency if it's not there yet
2. remove the base path
3. each test do the following:
  a. `@Before`: create a mock server and start it
  b. `mock.chain(request = <request you are expecting>, response = <response you want to>` 
  c. `Fuel.method(mock.path("the path"))`
  d. `@After`: shutdown mock server

You can usually use `mock.reflect()` which will simply give back the request with some body that holds what you sent.

lucasvalenteds:
Thank you for the feedback @kittinunf, @iNoles and @SleeplessByte. The changes were addressed in 205f18a and b927e40.

Let me know if there is anything else.

kittinunf:
Looks great for me.

PR #426, opened 2018-09-09
BREAKING CHANGE - Convert Base64 to Kotlin

iNoles:
There are no line terminators basically Android's Base64.NO_WRAP. 

Some developers would love extensions function that will make changing easily.

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/util/Base64.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/util/Base64.kt
@@ -19,13 +19,13 @@ fun ByteArray.encodeBase64(): ByteArray {
     var padding = 0
     var position = 0
     while (position < this.size) {
-        var b = this[position].toInt() and 0xFF shl 16 and 0xFFFFFF
-        if (position + 1 < this.size) b = b or (this[position + 1].toInt() and 0xFF shl 8) else padding++
-        if (position + 2 < this.size) b = b or (this[position + 2].toInt() and 0xFF) else padding++
+        var byte = this[position].toInt() and 0xFF shl 16 and 0xFFFFFF
+        byte = if (position + 1 < this.size) byte or (this[position + 1].toInt() and 0xFF shl 8) else padding++
+        if (position + 2 < this.size) byte = byte or (this[position + 2].toInt() and 0xFF) else padding++
         for (i in 0 until 4 - padding) {
-            val c = b and 0xFC0000 shr 18
+            val c = byte and 0xFC0000 shr 18
             output.write(table[c].toInt())
-            b = b shl 6
+            byte = byte shl 6
         }
         position += 3
     }
@@ -69,8 +69,8 @@ fun ByteArray.decodeBase64(): ByteArray {
             count++
         }
         while (count > 0) {
-            val c = b and 0xFF0000 shr 16
-            output.write(c.toChar().toInt())
+            val character = b and 0xFF0000 shr 16
+            output.write(character.toChar().toInt())
             b = b shl 8
             count--
         }


fuel/src/main/kotlin/com/github/kittinunf/fuel/util/Base64.kt
```diff
@@ -0,0 +1,80 @@
...
+            -1, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35, 36, 37, 38, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, -1, -1, -1, -1, -1, -1,
+            -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+            -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+            -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+            -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1)
+
+    val output = ByteArrayOutputStream()
+    var position = 0
+    while (position < this.size) {
+        var b: Int
```
markGilchrist: would it be better to use a more descriptive name like byte?

fuel/src/main/kotlin/com/github/kittinunf/fuel/util/Base64.kt
```diff
@@ -0,0 +1,80 @@
...
+        if (position + 2 < this.size && table[this[position + 2].toInt()] != -1) {
+            b = b or (table[this[position + 2].toInt()] and 0xFF shl 6)
+            count++
+        }
+        if (position + 3 < this.size && table[this[position + 3].toInt()] != -1) {
+            b = b or (table[this[position + 3].toInt()] and 0xFF)
+            count++
+        }
+        while (count > 0) {
+            val c = b and 0xFF0000 shr 16
```
markGilchrist: would `c` be better as a `character`?

fuel/src/main/kotlin/com/github/kittinunf/fuel/util/Base64.kt
```diff
@@ -0,0 +1,80 @@
...
+
+fun ByteArray.encodeBase64(): ByteArray {
+    val table = (CharRange('A', 'Z') + CharRange('a', 'z') + CharRange('0', '9') + '+' + '/').toCharArray()
+    val output = ByteArrayOutputStream()
+    var padding = 0
+    var position = 0
+    while (position < this.size) {
+        var b = this[position].toInt() and 0xFF shl 16 and 0xFFFFFF
+        if (position + 1 < this.size) b = b or (this[position + 1].toInt() and 0xFF shl 8) else padding++
+        if (position + 2 < this.size) b = b or (this[position + 2].toInt() and 0xFF) else padding++
```
markGilchrist: would this be clearer if this was written as 
```kotlin 
b = if (position + 2 < this.size) b or (this[position + 2].toInt() and 0xFF) else padding++
```

fuel/src/main/kotlin/com/github/kittinunf/fuel/util/Base64.kt
iNoles: You want to rename b to byte?

fuel/src/main/kotlin/com/github/kittinunf/fuel/util/Base64.kt
iNoles: I see what you meant.

fuel/src/main/kotlin/com/github/kittinunf/fuel/util/Base64.kt
SleeplessByte: Yes, it's better to use a descriptive name. 

Isn't it `decodedByte` or lookupByte or something like that?

iNoles:
@kittinunf I could try to cover all possibles test case from original test. Only behavior changed is no line terminators.

iNoles:
@kittinunf it may be good idea for testing original behaviors. This Kotlin version has no flags and line terminators.

fuel/src/main/kotlin/com/github/kittinunf/fuel/util/Base64.kt
markGilchrist: @iNoles  sorry I didn't reply to this sooner, yes please

fuel/src/main/kotlin/com/github/kittinunf/fuel/util/Base64.kt
```diff
@@ -0,0 +1,80 @@
...
+        position += 3
+    }
+    for (i in 0 until padding) {
+        output.write('='.toInt())
+    }
+    return output.toByteArray()
+}
+
+fun ByteArray.decodeBase64(): ByteArray {
+    val table = intArrayOf(-1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
```
markGilchrist: would this table not be better in a companion object or declared as a const val, as everytime this function is called it will have to declare this array to then junk it?

iNoles: I can't do const val because int[] is primitive types. 
I had 

```kotlin
val decodedTable = intArrayOf()
``` 
outside of the function

Inside of the decodeBase64 function, 
I had
```kotlin
 val table = decodedTable
```

iNoles:
I am thinking about to close this PR for different kinds of Base64.

kittinunf:
Do you have other Base64 implementation in mind? @iNoles

iNoles:
@kittinunf Yes, I do. It is in https://github.com/iNoles/Fuel/blob/Kotlin2Base64/fuel/src/main/kotlin/com/github/kittinunf/fuel/util/Base64.kt

kittinunf:
Let's do it then 👏

PR #431, opened 2018-09-14
Some Async Coroutine Functions

### LINKED ISSUE #429: Actually asynchronous coroutines
Right now, the the functions offered by `fuel-coroutines` simply start a coroutine which blocks until the request is complete. Ideally, this should actually suspend the coroutine instead of blocking whenever possible.

iNoles:
Thank you for submitting your Pull Request. Please make sure you have
familiarised yourself with the [Contributing Guidelines](https://github.com/kittunf/Fuel/CONTRIBUTING.md)
before continuing.

## Description

I added async block for some functions and left blocking one alone. I was a bit curious why somebody else didn't added this one. 

Fixes #429 

## Type of change

Check all that apply

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

tests for your changes, you may ommit this section.

## Checklist:


SleeplessByte:
I'm personally not a fan of the double await. asyncObjectResponse().await() would work, and allows you to have this be additive and not breaking.

iNoles:
@SleeplessByte Yeah, I see what you mean. you want to add extra functions for async*?

fuel-coroutines/src/main/kotlin/com/github/kittinunf/fuel/coroutines/Coroutines.kt
```diff
@@ -7,9 +7,7 @@ import com.github.kittinunf.fuel.core.Response
 import com.github.kittinunf.fuel.core.ResponseDeserializable
 import com.github.kittinunf.fuel.core.response
 import com.github.kittinunf.result.Result
-import kotlinx.coroutines.experimental.CommonPool
-import kotlinx.coroutines.experimental.suspendCancellableCoroutine
-import kotlinx.coroutines.experimental.withContext
+import kotlinx.coroutines.experimental.*
```
markGilchrist: would you mind reverting this change as we try not to use wildcard imports

fuel-coroutines/src/main/kotlin/com/github/kittinunf/fuel/coroutines/Coroutines.kt
```diff
@@ -23,22 +21,43 @@ private suspend fun <T : Any, U : Deserializable<T>> Request.await(
...
         }
 
-
 suspend fun Request.awaitByteArrayResponse(
         scope: CoroutineContext = CommonPool
-): Triple<Request, Response, Result<ByteArray, FuelError>> =
-        await(byteArrayDeserializer(), scope)
+): Triple<Request, Response, Result<ByteArray, FuelError>> = await(byteArrayDeserializer(), scope)
+
+suspend fun Request.asyncByteArrayResponse(
```
markGilchrist: is there a  reason for changing the name of this function?

iNoles: I didn't change it.

fuel-coroutines/src/main/kotlin/com/github/kittinunf/fuel/coroutines/Coroutines.kt
iNoles: Okay, I will.

fuel-coroutines/src/main/kotlin/com/github/kittinunf/fuel/coroutines/Coroutines.kt
markGilchrist: sorry I see that

fuel-coroutines/src/test/kotlin/com/github/kittinunf/fuel/coroutines/CoroutinesTest.kt
```diff
@@ -23,7 +23,7 @@ class CoroutinesTest {
     @Test
     fun testAwaitResponseSuccess() = runBlocking {
         try {
-            Fuel.get("/ip").awaitByteArrayResponse().third.fold({ data ->
```
kittinunf: Don't we want to keep `awaitXX` version for our user to use? Are we gonna deprecated them and move user to use the `asyncXX` version instead? I think either way we probably want to keep the tests for `awaitXXX`

SleeplessByte: Yeah definitely keep the tests until the await versions are removed, if ever.

iNoles: Oh Okay, I will restore awaitXX versions of tests.

kittinunf:
I think this looks great. It almost like we should deprecate the `awaitXX` version and expose only `asyncXX`. Because it looks like expose it as a `Deferred` triumphs other versions in every way 💦

all in all, after this PR is merged our Coroutine API landscape will look like this.

|Type\Return| * | Result<*, FuelError> | `Triple<Request, Response, Result<*, FuelError>`  | `Deferred<Triple<Request, Response, Result<*, FuelError>>>` |
|--| -- | -- | ------------- | ------------- |
|ByteArray| `awaitByteArray` | `awaitByteArrayResult ` |`awaitByteArrayResponse` | `asyncByteArrayResponse` |
|String| `awaitString` | `awaitStringResult ` |`awaitStringResponse`  | `asyncStringResponse`  |
|T| `awaitObject` | `awaitStringResult ` | `awaitObjectResponse` | `asyncObjectResponse` |

I think it might be "good" to provide the same treatment (provide the Deferred<> version) with the rest of the available API. 

But this means that we are going to double the available API for our users. ....

I almost feel like at this point, we might wanna clean this all up at once. It might be just wiser to return the Deferred version instead? 

What do you think?

iNoles:
@kittinunf I would love to deprecated Result and Triple and only uses Deferred.

SleeplessByte:
I concur with Jonathan: Deprecate, don't remove, but remove from
documentation.

kittinunf:
Just to confirm, I guess we want to go with just 2 versions!

|Type\Return| Deferred<Result<*, FuelError>> | Deferred<Triple<Request, Response, Result<*, FuelError>>> |
|--| -- | ------------- |
|ByteArray| `asyncByteArrayResult ` | `asyncByteArrayResponse` |
|String| `asyncStringResult ` | `asyncStringResponse`  |
|T| `asyncObjectResult ` | `asyncObjectResponse` |

And the rest, we deprecated them?

kittinunf:
Hmm, in my experience, `Result<*, FuelError>` is actually being used around 80% of my time, the rest about 20%, I need to check for HTTP status code, that is where the `Triple` version becomes handy. 

@iNoles If you don't mind, could you help me out?

iNoles:
@kittinunf Sure
I would go this route
```kotlin
fun awaitByteArrayResult(): Result<ByteArray, FuelError> = asyncByteArrayResponse().await().third
```

kittinunf:
Hmm, I think it is supposed to be `Deferred<Result<T, FuelError>>` ...

SleeplessByte:
The await version is not deferred because doing an . await blocks.

So the async one is deferred. Seems correct?

iNoles:
@SleeplessByte Deferred's await() will wait for Deferred to complete. 

@kittinunf I have a concern about it over asyncXXXResult because it is adding more complexity to the developers. It is going to be like  asyncXXXResponse().await().third.await(). It going to be Confusing for double await().

SleeplessByte:
@iNoles exactly what I thought :) And I agree with the double await being confusing.

clarcharr:
Using a `Deferred` doesn't actually solve this problem, sadly. It also seems like an anti-pattern because it's easy to just wrap the `await` in your own `async` call, whereas this forces you to use the `Deferred`, which is dispatched on a context you can't control.

The main problem described in #429 is that the coroutine still blocks instead of suspending the thread when it's waiting on I/O. This doesn't actually fix that problem.

iNoles:
> Using a `Deferred` doesn't actually solve this problem, sadly. It also seems like an anti-pattern because it's easy to just wrap the `await` in your own `async` call, whereas this forces you to use the `Deferred`, which is dispatched on a context you can't control.
> 
> The main problem described in #429 is that the coroutine still blocks instead of suspending the thread when it's waiting on I/O. This doesn't actually fix that problem.

Oh, what does solve this problem?

kittinunf:
Ahhhh sorry, I understand it now. Then, the current API proposal is OK to me.

@clarcharr I think we need you to perhaps shed some light here. I think the last thing we want is merging this and it doesn't solve your problem.

iNoles:
@kittinunf Do you want me to restore

```
fun awaitByteArrayResult(): Result<ByteArray, FuelError> = asyncByteArrayResponse().await().third
```
because 80% of the time?

kittinunf:
I feel like i am gonna use the coroutine version, I think I might prefer to use `Deferred<Result<T, E>` version. But because I don't use coroutine on my work at all so please take my opinion for a gain of salt.

clarcharr:
I'm not going to be able to get to this today, but I'll try this weekend to draft up the sort of changes that would be needed to fix the original issue. Part of it can be done without really affecting the surface API, but another part can't.

Basically, I think that establishing a connection to the server and getting back the response can be done asynchronously, where the only difference from the existing implementation is that the current thread isn't blocked from performing other work while the connection happens.

Making the read of the response body asynchronous would require changing the deserialization API, which would be breaking. But as far as things go, a majority of the time for the request is in getting back the response headers, which means that there's still a lot to gain without breaking the API.

markGilchrist:
@SleeplessByte @iNoles @clarcharr  I would really value your critique on my pr https://github.com/kittinunf/Fuel/pull/438

lucasvalenteds:
Just a note about returning a `Deferred<T>`: https://stackoverflow.com/questions/49762153/modify-the-deferred-result/49762498#49762498

iNoles:
I am going to close this  PR because async-await is BEST for Multiple coroutines with concurrency. Since 80% of the cases, it going to be single operation that is best for withContext.

PR #438, opened 2018-09-21
Add suspending implenetation client

markGilchrist:
## Description

In this pr I am hoping to create a suspending implementation of the httpClient so that, should a user so wish they could avoid using threads. 

Huge `thank you`/`love`/`creative respect`/`kudos` must go to both @iNoles, @lucasvalenteds and @SleeplessByte  who all helped me get this running at points I was completely stuck

Fixes #431 

Check all that apply

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## How Has This Been Tested?

This pr changes the implementation but the test remain the same as the object was not change the behaviour

## Checklist:


iNoles:
It does look good to me.

kittinunf:
@SleeplessByte Not related to this PR. But the content in the current PR template is more like a "placeholder"? Maybe we add ">" to make it looks like so? What do you think?

SleeplessByte:
@kittinunf sure, we can try out whatever we'd like. We can also add `<!-- comments -->` so if people don't change it stays empty!

iNoles:
@SleeplessByte I would like to add "Did you checked these files by the Kotlin ByteCode Viewer?" too

kittinunf:
I did share my concern already with @markGilchrist personally. But I would love to hear from others as well. With this PR, it means that we have push the coroutine down to `fuel` core which make me wonder whether this is the right apporach. At least for non co-routine user (like me), it means that we force all of our users to use

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -67,11 +68,62 @@ internal class HttpClient(private val proxy: Proxy? = null) : Client {
         }
     }
 
+    override suspend fun awaitRequest(request: Request): Response = suspendCoroutine { continuation ->
+        try {
```
kittinunf: Since this implementation is basically the same with `executeRequest` counterpart. Shall we abstract them out so we have a better maintainability later, even better if we can call the `executeRequest` directly?

``` Kotlin
override suspend fun awaitRequest(request: Request): Response = suspendCoroutine {
  try {
    continuation.resume(executeRequest(request))
  } catch(exception: Exception) {
    continuation.resumeWithException(exception)
  }
}

```

SleeplessByte:
I think we can keep @markGilchrist his implementation but NOT push this into core by actually extend the fuel manager's client initialisation to use an async client (as implemented here) when use fuel-coroutines.

Yes, this means one extra "setup" method call for the user, but I think it is well worth it to keep this out of core and keep it isolated in coroutines @kittinunf .

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -67,11 +68,62 @@ internal class HttpClient(private val proxy: Proxy? = null) : Client {
...
+                        if (contentEncoding.compareTo("gzip", true) == 0) GZIPInputStream(stream) else stream
+                    } catch (exception: IOException) {
+                        try {
+                            connection.errorStream ?: connection.inputStream?.close()
+                        } catch (exception: IOException) {
+                        }
+                        ByteArrayInputStream(ByteArray(0))
+                    }
+            ))
+        } catch (exception: Exception) {
```
lucasvalenteds: Do we have tests to all possible exceptions that can be thrown here?

iNoles: By looking https://codecov.io/gh/kittinunf/Fuel/src/master/fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt, not much.

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
markGilchrist: yes, sorry I should have seen that first time round

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
markGilchrist: @lucasvalenteds sorry if I have missed a step here but I thought that all exceptions would be caught in the block as they all inherit from Exception?

I am sorry if I have misunderstood the problem

lucasvalenteds: @markGilchrist The question is if there are tests that verify the subtypes. For exemple, if a function inside the `try {}` block can throw a `IllegalArgumentException`, is there a test that demonstrate that happening?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Client.kt
```diff
@@ -1,5 +1,9 @@
 package com.github.kittinunf.fuel.core
 
+import com.github.kittinunf.result.Result
```
iNoles: I think this import should be removed.

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
markGilchrist: I have made this change but for some reason it is not showing

iNoles:
When I replaced
```
fail("When using awaitSafelyObjectResult errors should be folded instead of thrown."
```
to
```
fail(exception.localizedMessage)
```
I get com.fasterxml.jackson.module.kotlin.MissingKotlinParameterException cannot be cast to com.github.kittinunf.fuel.core.FuelError

iNoles:
About testAwaitForObjectResultCatchesDeserializeError, it seems your code is not really catching all errors.

iNoles:
```
 (state as? CompletedExceptionally)?.let { throw it.cause }
```
this actually bypass any try/catch code

iNoles:
@markGilchrist When I took some of your code from the scratch, I got it to work.
Do you want me to release it?

markGilchrist:
@iNoles if you want to raise another or tell me how to fix this one please go for whichever suits you I am 100% easy either way

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/TaskRequest.kt
```diff
@@ -26,16 +26,17 @@ internal open class TaskRequest(internal val request: Request) : Callable<Respon
     }
 
     open suspend fun awaitResult(): Result<Response, FuelError> {
```
iNoles: Is this ever uses anymore?

kittinunf:
I think we can target 1.3 by using https://dl.bintray.com/kotlin/kotlin-eap and `1.3.0-rc-57` as version number so that we can see how this can be when used with 1.3.0. @markGilchrist Do you think we should just do it in this PR?

markGilchrist:
@kittinunf I am biased but I feel we should merge it, however I am happy to wait until we are using 1.3 and it has been widely adopted.

I have removed (locally) 

``` kotlin
configure<KotlinJvmProjectExtension> {
            experimental.coroutines = Coroutines.ENABLE
        }

```

it seams to have very little effect other than it adds a compiler warning to the code.

I am aware that there may be issues that are beyond my comprehension and I am happy to do whatever the team feel is best

iNoles:
It does looks good in despite of jacoco errors. 

With Kotlin 1.3, that requires to change many things such as imports and method signature.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/AsyncTaskRequest.kt
```diff
@@ -22,4 +24,5 @@ internal class AsyncTaskRequest(private val task: TaskRequest) : TaskRequest(tas
     }
 
     private fun errorResponse() = Response(request.url)
-}
\ No newline at end of file
+}
+
```
kittinunf: 👮

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/AsyncTaskRequest.kt
```diff
@@ -2,6 +2,8 @@ package com.github.kittinunf.fuel.core.requests
 
 import com.github.kittinunf.fuel.core.FuelError
 import com.github.kittinunf.fuel.core.Response
+import com.github.kittinunf.result.Result
+
```
kittinunf: 👮

fuel-coroutines/src/test/kotlin/com/github/kittinunf/fuel/coroutines/CoroutinesTest.kt
```diff
@@ -10,7 +9,7 @@ import org.junit.Assert.*
 import org.junit.Before
 import org.junit.Test
 import java.net.HttpURLConnection
-import java.util.UUID
+import java.util.*
```
kittinunf: 👮

SleeplessByte:
Linking here for reference: https://www.youtube.com/watch?v=_hfBv0a09Jc, https://github.com/Kotlin/kotlin-coroutines/blob/master/kotlin-coroutines-informal.md#asynchronous-programming-styles

I'll be watching and reading and then re-reviewing.

build.gradle.kts
```diff
@@ -8,17 +8,22 @@ import org.jetbrains.kotlin.gradle.dsl.KotlinProjectExtension
...
+        maven(url = "https://dl.bintray.com/kotlin/kotlin-eap")
+        maven { setUrl("http://dl.bintray.com/kotlin/kotlin-eap") }
     }
 
     dependencies {
         classpath(Plugins.android)
         classpath(Plugins.jacocoAndroid)
         classpath(Plugins.bintray)
         classpath(kotlin("gradle-plugin", version = Versions.kotlinVersion))
+        classpath(kotlinModule("gradle-plugin", kotlin_version))
```
iNoles: Why we have two kotlin functions here?

fuel-android/build.gradle.kts
```diff
@@ -1,7 +1,16 @@
+val kotlin_version: String by extra
 dependencies {
     api(project(":fuel"))
     implementation(Dependencies.kotlinStdlib)
     testImplementation(Dependencies.robolectric)
     testImplementation(Dependencies.junit)
     testCompile(Dependencies.mockServer)
+    compile(kotlinModule("stdlib-jdk7", kotlin_version))
```
iNoles: oh course, import org.jetbrains.kotlin.config.KotlinCompilerVersion 
you can use kotlinModule("stdlib-jdk7", KotlinCompilerVersion.VERSION)

fuel-forge/build.gradle.kts
```diff
@@ -1,7 +1,24 @@
-plugins { java }
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins { java
+    kotlin("jvm") version "1.3.0-rc-116"
```
iNoles: you don't need version one here because you already have gradle-plugin on the classpath.

iNoles:
in root/build.gradle.kts, there is
```
compile(Dependencies.kotlinStdlib)
```

build.gradle.kts
markGilchrist: this was generated by the IDE

markGilchrist:
@lucasvalenteds I am having trouble setting the language to the latest version of kotlin (1.3.0-rc-116) can you see where I may be going wrong, any help here really appreciated!

lucasvalenteds:
@markGilchrist [here](https://github.com/lucasvalenteds/Fuel/commit/2c93f35dd2855eef17e4b8942182400fe28e82cf) is the code to migrate from Kotlin 1.2 to 1.3-M2. Running `./gradlew clean test` using OpenJDK 1.8 should work and print some deprecation warning.

Send me a message on Slack if you need anything. :+1:

iNoles:
@lucasvalenteds Are you talking about Kotlin's slack channel?

lucasvalenteds:
@iNoles Yes, [this one](https://kotlinlang.slack.com).

iNoles:
@lucasvalenteds I usually to be on that channel for some times.

By looking to https://github.com/gradle/kotlin-dsl/blob/master/buildSrc/src/main/kotlin/codegen/GenerateKotlinDependencyExtensions.kt, I have seen WHY IDE picked kotlinModule() over kotlin()

iNoles:
Gotta love those Samples. :|

iNoles:
@markGilchrist https://github.com/kittinunf/Fuel/blob/master/sample/src/main/kotlin/com/example/fuel/MainActivity.kt

fuel-coroutines/build.gradle.kts
```diff
@@ -1,16 +1,13 @@
 import com.novoda.gradle.release.PublishExtension
 import org.jetbrains.kotlin.gradle.dsl.Coroutines
 import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
-import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
 
-plugins { java }
+plugins { java
+    kotlin("jvm")
```
kittinunf: I think kotlin plugin is already applied at the `build.gradle` at the root level? I don't think we need it here?

fuel-coroutines/src/test/kotlin/com/github/kittinunf/fuel/coroutines/CoroutinesTest.kt
```diff
@@ -1,21 +1,20 @@
...
 import java.util.regex.Pattern
 
 class CoroutinesTest {
 
-    val uuidRegex = Pattern.compile("^[A-Fa-f0-9]{8}-[A-Fa-f0-9]{4}-4[A-Fa-f0-9]{3}-[A-Fa-f0-9]{4}-[A-Fa-f0-9]{12}$")
+    private val uuidRegex: Pattern = Pattern.compile("^[A-Fa-f0-9]{8}-[A-Fa-f0-9]{4}-4[A-Fa-f0-9]{3}-[A-Fa-f0-9]{4}-[A-Fa-f0-9]{12}$")
```
kittinunf: Not sure why we need to explicitly state type here?

fuel-coroutines/src/test/kotlin/com/github/kittinunf/fuel/coroutines/CoroutinesTest.kt
```diff
@@ -39,8 +38,8 @@ class CoroutinesTest {
     @Test
     fun testAwaitResponseSuccess() = runBlocking {
         mock.chain(
-            request = mock.request().withPath("/ip"),
-            response = mock.reflect()
+                request = mock.request().withPath("/ip"),
```
kittinunf: [Nits] Not sure about these indentations though? Have you tried to use the auto format from Intellij/AS?

fuel-coroutines/src/test/kotlin/com/github/kittinunf/fuel/coroutines/CoroutinesTest.kt
kittinunf: 🙏

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/RedirectionInterceptor.kt
```diff
@@ -1,11 +1,6 @@
 package com.github.kittinunf.fuel.core.interceptors
 
-import com.github.kittinunf.fuel.core.Encoding
-import com.github.kittinunf.fuel.core.FuelManager
-import com.github.kittinunf.fuel.core.Method
-import com.github.kittinunf.fuel.core.Request
-import com.github.kittinunf.fuel.core.Response
-import com.github.kittinunf.fuel.core.isStatusRedirection
+import com.github.kittinunf.fuel.core.*
```
kittinunf: 👮‍♀️

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/SuspendingRequest.kt
```diff
@@ -0,0 +1,28 @@
...
+
+
+import com.github.kittinunf.fuel.core.FuelError
+import com.github.kittinunf.fuel.core.Request
+import com.github.kittinunf.fuel.core.Response
+import com.github.kittinunf.result.Result
+import com.github.kittinunf.result.mapError
+import java.io.InterruptedIOException
+
+class SuspendingRequest(internal val request: Request) {
```
kittinunf: 🤔 Why `internal`?

kittinunf:
I have been sitting in a lot of talks about Coroutine today at Kotlin Conf. (apparently, it is a very hot topic ofc) and I realize that the correct usage for a client to use Fuel should look something like at the client codebase.

Definition
``` Kotlin
interface RemoteService {
  suspend fun get(): Result<String, FuelError>
}
```
Usage
``` Kotlin
val result = withContext(dispatchers.io) {
  service.getFoo()
} 

when (result) {
  is Sucess -> { //do something with success }
  is Failure -> { //do something with failure }
}
```

Then, for `Fuel` to supply the implementation. It probably should **be** something along this line


``` Kotlin
class RemoteServiceImpl : RemoteService { 
  override suspend fun get(): Result<String, FuelError> = Fuel.get("...").awaitStringResult()
}
```

I think we might add this into Readme to instruct user! As long as we are able to do this in this PR, it looks like we are going on the right path 🎉 👍

iNoles:
so it is like Retrofit, right?

fuel-coroutines/src/test/kotlin/com/github/kittinunf/fuel/coroutines/CoroutinesTest.kt
markGilchrist: that is what I am using

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/SuspendingRequest.kt
markGilchrist: I have made it private as access is not currently required outside the class

kittinunf:
@iNoles we don't care about Retrofit here :smi

kittinunf:
@markGilchrist one last thing

In this PR we change our core interface into 

``` Kotlin
interface Client {
  fun executeRequest(request: Request): Response
  suspend fun awaitRequest(request: Request): Response
} 
```

What do you think what the concrete class that doesn't care about coroutine is going to be? For example, `TestClient` in for testing purposes?

@iNoles 
Using interface to back the concrete class is a reasonable practice anyway 😏

markGilchrist:
@kittinunf

Hmm well here is my two cents

## The argument for keeping this as it is

I feel that the Client is the thing that delivers the request, and I feel that you should in that case any class that implements this interface should provide both implementations.

I don't feel this adds a problem to the tests, as can be seen in `CustomClientTest.kt` line 28. perhaps I could add a default implementation that simply calls the non suspending implementation? 

```kotlin
override suspend fun awaitRequest(request: Request) = executeRequest(request)
```

## An other route?

Currently all requests have a client, if you were to separate out the `Client` to lets say `SuspendingClient` and `Client` then you have in my opinion a problem as you will then have to add  another property to Request and you would have to add logic to ensure that the Request has at least on client.

as ever your wish is my command

kittinunf:
I think the retry feature is killing us on the PR build lol

kittinunf:
I would love to hear some other opinions from other, however I think this PR is already good as it is. I will give this a go but I still believe that we need to do "something" so that the `interface Client` is simple to be used.

markGilchrist:
Are we going to wait until the main branch is using 1.3?

build.gradle.kts
```diff
@@ -4,22 +4,25 @@ import com.jfrog.bintray.gradle.BintrayExtension
...
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 buildscript {
     repositories {
         mavenCentral()
         jcenter()
         google()
+        maven(url = "https://dl.bintray.com/kotlin/kotlin-dev")
         maven { setUrl("https://kotlin.bintray.com/kotlinx") }
+        maven { setUrl("http://dl.bintray.com/kotlin/kotlin-eap") }
```
SleeplessByte: Probably all should use `url = `

fuel-coroutines/src/test/kotlin/com/github/kittinunf/fuel/coroutines/CoroutinesTest.kt
SleeplessByte: Ha seems to be back :P

fuel-coroutines/src/test/kotlin/com/github/kittinunf/fuel/coroutines/CoroutinesTest.kt
SleeplessByte: The auto format isn't completely sane by default. It will "over indent" function parameters on a new line. Perhaps one of us should toy with the settings, and then make the settings available (you can export/import them).

iNoles:
@SleeplessByte it does work only on test failure but not gradle dependency.  :(

build.gradle.kts
markGilchrist: I have made this change everywhere in the code base

markGilchrist:
@iNoles  which test is failing?

markGilchrist:
@iNoles if I have understood this properly https://github.com/Kotlin/kotlinx.serialization/issues/152 there is a bug in the kotlinx version that I am using. I am not sure this can be fixed atm

iNoles:
@markGilchrist you could change the serialization version to 0.8.2-rc13 ONLY for runtime, but not classpath one.

markGilchrist:
@iNoles  you are right!

iNoles:
I see Two warnings. 
1) 'typealias KSerialLoader<T> = DeserializationStrategy<T>' is deprecated. Obsolete name from the preview version of library.'

2) Incompatible version of kotlin-reflect in fuel-reactors

kittinunf:
I think I am good to merge this in. Shall we merge it and release new version?

markGilchrist:
@kittinunf  I think we can merge it in ?

SleeplessByte:
> I see Two warnings.
> 
> 1. 'typealias KSerialLoader = DeserializationStrategy' is deprecated. Obsolete name from the preview version of library.'
> 2. Incompatible version of kotlin-reflect in fuel-reactors

@iNoles this can be fixed in a subsequent PR I suppose, as it doesn't really have anything todo with this?

NikkyAI:
i will do the PR to fix kotlinx.serialization stuff and deprecations, `0.6.2` -> `0.8.0`
there have been a lot of renames and probably better ways to do the things

clarcharr:
So erm, I never got around to taking a look at this, but is there a particular reason why the language version was bumped to 1.3? This completely breaks compatibility with 1.2, and I don't think that any of the features used actually require 1.3.

PR #443, opened 2018-09-24
Fixed and Improve Coverage of Forge

iNoles:
Thank you for submitting your Pull Request. Please make sure you have
familiarised yourself with the [Contributing Guidelines](https://github.com/kittunf/Fuel/CONTRIBUTING.md)
before continuing.

## Description

This will improve the Test Coverage of Forge.

Fixes #0
Fixes #0

## Type of change

Check all that apply

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

tests for your changes, you may ommit this section.

## Checklist:


fuel-forge/src/main/java/com/github/kittinunf/fuel/forge/FuelForge.kt
```diff
@@ -22,14 +22,9 @@ inline fun <reified T : Any> Request.responseObject(noinline deserializer: JSON.
 
 inline fun <reified T : Any> Request.responseObjects(noinline deserializer: JSON.() -> DeserializedResult<T>) = response(forgesDeserializerOf(deserializer))
 
-inline fun <reified T : Any> forgeDeserializerOf(noinline deserializer: JSON.() -> DeserializedResult<T>) = object : ResponseDeserializable<T> {
-
+fun <T : Any> forgeDeserializerOf(deserializer: JSON.() -> DeserializedResult<T>) = object : ResponseDeserializable<T> {
```
kittinunf: Maybe we mark this as private as we don't want our client to call this directly.

fuel-forge/src/main/java/com/github/kittinunf/fuel/forge/FuelForge.kt
```diff
@@ -22,14 +22,9 @@ inline fun <reified T : Any> Request.responseObject(noinline deserializer: JSON.
...
     override fun deserialize(content: String): T? = Forge.modelFromJson(content, deserializer).component1()
-
 }
-
-inline fun <reified T : Any> forgesDeserializerOf(noinline deserializer: JSON.() -> DeserializedResult<T>) = object : ResponseDeserializable<List<T>> {
-
+fun <T : Any> forgesDeserializerOf(deserializer: JSON.() -> DeserializedResult<T>) = object : ResponseDeserializable<List<T>> {
```
kittinunf: Here as well

fuel-forge/src/main/java/com/github/kittinunf/fuel/forge/FuelForge.kt
iNoles: If I marked it as private, I think I will break the test.

markGilchrist: This does feel like we have removed the `inline` modifier which will remove any performance enhancements and it makes the jacoco diff look better, however if this pr contains no more unit test or assertions, in which case how have we really improved the coverage?

iNoles:
@markGilchrist I think by removing inline, it will reduce a size of kotlin bytecode level. With inline, it added extra $inline$ classes. Without inline, it will be into function itself in the same class.

SleeplessByte:
As per the [kotlin documentation](https://kotlinlang.org/docs/reference/inline-functions.html):
 
> Inlining may cause the generated code to grow; however, if we do it in a reasonable way (i.e. avoiding inlining large functions), it will pay off in performance, especially at "megamorphic" call-sites inside loops.

So small functions *should* be inlined, because it speeds them up (no more >polymorphic calls, less memory), despite them growing byte code. 

Is there a way to measure that reliably? Can we add a memory benchmark (not running each test, but for us) that has some reasonable usage of this function so we can actually see the difference in size + difference in memory?

kittinunf:
I also do not feel that changing from inline -> not inline make a lot of differences, unless we have the proof to say otherwise.

iNoles:
Hmm, this only increased 40% of the FuelForge files.

PR #444, opened 2018-09-25
Mock all the projects

SleeplessByte:
## Description

Based on #434, this PR brings mocking to *all* the other modules, which should:
- speed up the entire suite
- get more consistent test results
- work offline
- no more failures because github.com or httpsbin are failing.

The only "downside" of this is the duplication of `MockHelper`. If someone knows how to fix up the gradle config so we can import the one from `fuel` core everywhere... that'd be great.

## Type of change

Check all that apply

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

## Checklist:


fuel-android/src/test/kotlin/com/github/kittinunf/fuel/android/BaseTestCase.kt
```diff
@@ -1,21 +1,38 @@
...
     lateinit var lock: CountDownLatch
 
     fun await(seconds: Long = DEFAULT_TIMEOUT) {
         lock.await(seconds, TimeUnit.SECONDS)
     }
 
+
+    @Before
+    fun setup() {
+        this.mock = MockHelper()
```
markGilchrist: do you need `this`?

fuel-android/src/test/kotlin/com/github/kittinunf/fuel/android/RequestAndroidAsyncTest.kt
```diff
@@ -267,8 +332,8 @@ class RequestAndroidAsyncTest : BaseTestCase() {
         assertThat(res, notNullValue())
         assertThat(err, nullValue())
         assertThat(data, notNullValue())
-        assertThat(data as HttpBinHeadersModel, isA(HttpBinHeadersModel::class.java))
-        assertThat((data as HttpBinHeadersModel).headers.isNotEmpty(), isEqualTo(true))
+        assertThat(data!!.headers.isNotEmpty(), isEqualTo(true))
```
markGilchrist: if it worth putting these tests in 

``` kotlin
if (data != null){

}```

so we can remove the !!

File: fuel-coroutines/build.gradle.kts [hunk 3]
```diff
@@ -8,7 +8,7 @@ plugins { java }
 dependencies {
     compile(project(":fuel"))
     compile(Dependencies.kotlinCoroutinesJvm)
-    testCompile(project(":fuel-jackson"))
+    testCompile(Dependencies.mockServer)
```
markGilchrist: nice

fuel-coroutines/src/test/kotlin/com/github/kittinunf/fuel/coroutines/CoroutinesTest.kt
```diff
@@ -131,13 +192,18 @@ class CoroutinesTest {
...
+
         try {
-            Fuel.get("/ip").awaitObjectResult(UUIDResponseDeserializer).fold({ _ ->
+            Fuel.get(mock.path("ip")).awaitObjectResult(UUIDResponseDeserializer).fold({ _ ->
                 fail("This is an error case!")
 
             }, { error ->
                 assertNotNull(error)
-                assertTrue(error.exception is JsonMappingException)
+                assertTrue(error.exception is UUIDResponseDeserializer.NoValidFormat)
```
markGilchrist: could we make this an assertEqual? then you get a better explanation

fuel-android/src/test/kotlin/com/github/kittinunf/fuel/android/BaseTestCase.kt
SleeplessByte: Does it work without? Sorry I think I might be mixing languages here :). @markGilchrist what are the rules in Kotlin when you must prefix with this? Is it like ruby (only when you have a conflicting identifier, or a "private" setter?)

fuel-android/src/test/kotlin/com/github/kittinunf/fuel/android/RequestAndroidAsyncTest.kt
SleeplessByte: Line 334 will fail if it's null. We basically should add "assert data != null" at line 335 so the kotlin compiler knows that the code would have stopped if it's... null.

Wrapping just increases complexity whilst not solving anything because of line 334.

fuel-coroutines/src/test/kotlin/com/github/kittinunf/fuel/coroutines/CoroutinesTest.kt
SleeplessByte: Sure!

fuel-android/src/test/kotlin/com/github/kittinunf/fuel/android/BaseTestCase.kt
iNoles: https://kotlinlang.org/docs/reference/this-expressions.html It can work without it.

markGilchrist: @SleeplessByte  sorry for my late reply, you only need this if the compiler can't work out which variable you are taking about

```kotlin
fun setUser(user:User){
    this.user = user
    user = user // the compiler thinks you are assigning the variable to itself
}
```

fuel-android/src/test/kotlin/com/github/kittinunf/fuel/android/RequestAndroidAsyncTest.kt
markGilchrist: I guess you are asserting not null in the line above (334)

kittinunf:
I also really like this @SleeplessByte. Really thanks for your hard work! 

Just one small note, I see we are duplicate `class MockHelper` everywhere. Do you think we should have a share test module (maybe name `test-util`) and share the implementation with `testImplemenation`?

We can do this in a different PR too. I will give an approve signal to merge this.

fuel-android/src/test/kotlin/com/github/kittinunf/fuel/android/BaseTestCase.kt
SleeplessByte: Thanks! Same as in Ruby then, I can work with this :) (well without `this` :P)

fuel-android/src/test/kotlin/com/github/kittinunf/fuel/android/RequestAndroidAsyncTest.kt
SleeplessByte: Exactly, so the compiler doesn't understand the test assertion, but we know it will ALWAYS be not null if it passes the line above.

SleeplessByte:
@kittinunf Yes! I would like a test lib a lot, especially now. I would like to define a set of common requests and responses, so we can actually test ALL the libraries against the same requests, and then we have waaaay less duplication. 🗡 

Naturally you would then only tailor the deserialization and serialization etc, but that'll come.

PR #465, opened 2018-10-17
Replace buildscript with plugin DSL

NikkyAI:
## Description

Replaced `buildscript` block with `plugins`
this will allow gradle to `Optimize the loading and reuse of plugin classes.` (at least according to their [documentation](https://docs.gradle.org/current/userguide/plugins.html#plugins_dsl_limitations))

IT should be discussed where versions and ids should be defined,
in theory plugins could also be applied as extension values somewhere in `buildSrc`

I organized it i na way that loads potentially faster and makes the most sense to me for now

## Type of change

Check all that apply

- [x] Refactoring build script (a change which changes the current internal or external interface)

## How Has This Been Tested?

This should **NOT** affect any tests or the module structure, just reload the gradle nature of the project in idea as usual

## Checklist:

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -7,26 +7,14 @@ import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
 import org.jetbrains.kotlin.gradle.dsl.KotlinProjectExtension
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
-buildscript {
-    repositories {
-        mavenCentral()
-        jcenter()
-        google()
-        maven(url = "https://dl.bintray.com/kotlin/kotlin-dev")
-        maven(url = "https://kotlin.bintray.com/kotlinx")
-        maven(url = "http://dl.bintray.com/kotlin/kotlin-eap")
-    }
-
-    dependencies {
-        classpath(Plugins.android)
-        classpath(Plugins.jacocoAndroid)
-        classpath(Plugins.bintray)
-        classpath(Plugins.kotlin)
-        classpath(Plugins.serialization)
-    }
+plugins {
+    java
+    kotlin("jvm") version Versions.kotlinVersion apply false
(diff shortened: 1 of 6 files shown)


settings.gradle.kts
```diff
@@ -2,8 +2,29 @@ pluginManagement {
     repositories {
         maven(url = "http://dl.bintray.com/kotlin/kotlin-eap")
         mavenCentral()
+        jcenter()
+        google()
         maven(url = "https://plugins.gradle.org/m2/")
     }
+    resolutionStrategy {
```
kittinunf: Do we really need this?

build.gradle.kts
```diff
@@ -7,27 +7,16 @@ import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
...
-        classpath(Plugins.jacocoAndroid)
-        classpath(Plugins.bintray)
-        classpath(Plugins.kotlin)
-        classpath(Plugins.serialization)
-    }
+plugins {
+    java
+//    kotlin("jvm") version Versions.kotlinVersion apply false
+    id(PluginID.kotlinJvm) version Versions.kotlinVersion apply false
+    id(PluginID.androidLib) version "3.1.3" apply false
```
kittinunf: We probably want to extract them out as variables too like the kotlinVersion one.

settings.gradle.kts
iNoles: as for new plugins block, Yup.

build.gradle.kts
NikkyAI: done

settings.gradle.kts
NikkyAI: if google and jetbrains cared about it more they may put gradle plugin markersin the repo.. but well not yet..

NikkyAI:
i think this would be ready to merge

PR #474, opened 2018-10-18
Replace Content-Type when jsonBody is called

### LINKED ISSUE #473: jsonBody does not replace default content encoding
# Bug Report

## Description
when POST-ing a json body the generated request contains both of these headers
```
Content-Type : application/x-www-form-urlencoded
content-type : application/json
```

because of https://github.com/kittinunf/Fuel/blob/master/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt#L149

## To Reproduce

```Kotlin
val curl = Fuel.post("https://someurl.com")
            .jsonBody(body = "{}")
            .cUrlString()
println(curl)
```

## Expected behavior
the POST request should have only one Content-Type

NikkyAI:
## Description

Ensures that the default Content-Type of the request gets replaced instead of appending to the list or adding to a different header when `jsonBody` is called.

eventually case insensitive headers might be useful

Fixes #473

## Type of change

Check all that apply

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
@@ -146,7 +146,7 @@ class Request(
     fun body(body: String, charset: Charset = Charsets.UTF_8): Request = body(body.toByteArray(charset))
 
     fun jsonBody(body: String, charset: Charset = Charsets.UTF_8): Request {
-        header("content-type" to "application/json")
+        headers["Content-Type"] = "application/json"
         return body(body, charset)
     }
 
--- a/fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestTest.kt
+++ b/fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestTest.kt
@@ -270,7 +270,7 @@ class RequestTest : MockHttpTestCase() {
     fun httpPostRequestWithBody() {
         val foo = "foo"
         val bar = "bar"
-        val body = "{ $foo : $bar }"
+        val body = "{ $foo: $bar }"
 
         // Reflect encodes the body as a string, and gives back the body as a property of the body
         //  therefore the outer body here is the ey and the inner string is the actual body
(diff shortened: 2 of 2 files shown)


SleeplessByte:
This is a patch that will work, and as you already note headers **SHOULD** be normalised (and thus are case-insensitive).

I'm okay with merging this, but let's keep track of this issue and actually fix it down the line :)

SleeplessByte:
@NikkyAI if you want, could you add a test that FAILS with the old behaviour and PASSES with the new behaviour?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
```diff
@@ -146,7 +146,7 @@ class Request(
     fun body(body: String, charset: Charset = Charsets.UTF_8): Request = body(body.toByteArray(charset))
 
     fun jsonBody(body: String, charset: Charset = Charsets.UTF_8): Request {
-        header("content-type" to "application/json")
+        headers["Content-Type"] = "application/json"
         return body(body, charset)
     }
```
markGilchrist: should we amend the test for this `RequestTest.kt` line 301, currently the test passes and if it is a bug the test should also be changed to reflect the desired behaviour

SleeplessByte:
Seems to revert the constants stuff @NikkyAI

NikkyAI:
oops

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
kittinunf: Agree with that. So we have test that covered the behavior.

markGilchrist:
@kittinunf  @NikkyAI  have you updated the tests?

NikkyAI:
i was not aware there were tests to update or fix
.. and i am not sure how most of them work :/

markGilchrist:
@NikkyAI  if you `cmd` click on the function it will show you where it is called (Mac + Android Studio)

RequestTest.kt, line 301, I am guessing you need to update line 308

NikkyAI:
the tests currently fail already when there is mutliple fields in "Content-Type" the only thing left to cehck if someone accidentally used multiple different casings for the content-type header

NikkyAI:
Just tested this, the test also fails when there is expected/different content-type header with different casing
since the list of header will be different the test will catch it

since that seems safe, i think its ready to merge ?

Issue #263, opened 2017-10-31
CookieStore support?


storezhang:
When can you support cookie functionality?

PerfectedTech:
+1

markGilchrist:
@storezhang 
if you can elaborate on what functionality you are looking for I will make a pr and implement it

C06A:
Cookies transported between client and server as header.
The CookieStore is not part of the communication protocol and should be implemented by client you are building on top of Fuel. It is not that difficult.

The only functionality in support Cookies I see is parsing and formatting Cookies header in Response and Request respectively.

markGilchrist:
@C06A 

Thanks for your answer, does this pr -> https://github.com/kittinunf/Fuel/pull/393 go someway to starting this functionality?

Do you have a working example I can look at so I get a precise list of functionality to build this?

SleeplessByte:
@markGilchrist @kittinunf I think the request here is not specifically cookie store, but session management (which for a client is basically cookie persistence). On android, the cookie store here: https://developer.android.com/reference/java/net/CookieStore is automagically called! This is what calls it: https://developer.android.com/reference/java/net/CookieManager.

Additionally, the HttpClient used right now is explicitly set to NOT use caching, whereas `Android` has a great File cache that works similarly. Needs a separate ticket, but similar implementation. 

It probably should **NOT** be Fuel's responsibility to handle sessions, just allow for it (have the interface).

markGilchrist:
@SleeplessByte do we need to define our own interface or should we recommend users user the ones provided by android? do we need to implement any code or is this outside our remit for this project?

C06A:
Fuel is not implemented specifically for Android, so I would recommend to avoid platform-specific implementation.
As I mentioned before I think this is a functionality of the higher order.
Especially if we talking about session management. Client is not always browser and server may require different ways to communicate session info. All these depends on specific task.

For example REST API provided by Spring security relies on completely different set of headers and not cookies at all to manage sessions.

SleeplessByte:
> Fuel is not implemented specifically for Android, so I would recommend to avoid platform-specific implementation.
Agreed with this! @markGilchrist so your own interface would probably work well; or not do it at all but have access points so people can do it themselves.

The easiest solution is probably something along the lines of:
1. extract the current http client creation into a factory method that people can intercept
2. make it easy to set the http client / read it
3. add examples in the docs how to make this work with java 7/8 and android  

[...]

PR #481, opened 2018-10-25
Header normalisation and request fixes

### LINKED ISSUE #392: Parameters should be Map<String, Any?>
In my opinion, the `parameter` argument in the `httpGet()`,`httpPost()` ect. should be of type `Map<String, Any?>`. 

Kotlin allows you, to convert `List<Pair>>` to `Map` like this:

    
    val myMap = parameters.toMap()
    

Why not use `Map` instead of `List<Pair>`?

### LINKED ISSUE #408: change in behaviour of POST between 1.14.0 and 1.15.0
It seems a change was introduced sometime in 1.15.0

This used to work
``` kotlin
    fun getSessionInfo(tokenId: String): AmSession {
        logger.debug { "fetching info for session: $tokenId" }
        val (req, res, result) =
            http.request(
                method = Method.POST,
                path = "/json/sessions?_action=getSessionInfo&tokenId=$tokenId"
            )
                    .header("Accept-API-Version" to "resource=2.0")
                    .header("Content-Type" to "application/json")
                    .responseString()
        logger.debug { "\n${req.cUrlString()}\n--> $req\n$res" }
        result.fold({ d ->
            return Json().parse(d)
        }, { err ->
            throw err.exception
        })
    }
```

and it sent this request 

```
$ curl -i -X POST -H "Accept-Encoding:compress;q=0.5, gzip;q=1.0" -H "Content-Type:application/json" -H "Accept:application/json" -H "Accept-API-Version:resource=2.0" https://authserver/json/sessions?_action=getSessionInfo&tokenId=d5zqRsYnIPTqXvZRZGvdEdOB9Gs.*AAJTSQACMDIAAlNLABxPWFZLMHRhMHBHeXl6YjMrQ1h5Wk5LdGJubFk9AAJTMQACMDE.*
--> "Body : (empty)"
"Headers : (4)"
Accept-Encoding : compress;q=0.5, gzip;q=1.0
Content-Type : application/json
Accept : application/json
Accept-API-Version : resource=2.0
```

Notice the content type is "application/json" as is expected due to it being requested in code example on line 10

however in newer versions the request is this, notice the content type header being appended for some reason, instead of overwritting

```
$ curl -i -X POST -H "Accept-Encoding:compress;q=0.5, gzip;q=1.0" -H "Content-Type:application/x-www-form-urlencoded; application/json" -H "Accept:application/json" -H "Accept-API-Version:resource=2.0" https://authserver/json/sessions?_action=getSessionInfo&tokenId=d5zqRsYnIPTqXvZRZGvdEdOB9Gs.*AAJTSQACMDIAAlNLABxPWFZLMHRhMHBHeXl6YjMrQ1h5Wk5LdGJubFk9AAJTMQACMDE.*
--> --> https://authserver/json/sessions?_action=getSessionInfo&tokenId=d5zqRsYnIPTqXvZRZGvdEdOB9Gs.*AAJTSQACMDIAAlNLABxPWFZLMHRhMHBHeXl6YjMrQ1h5Wk5LdGJubFk9AAJTMQACMDE.*
"Body : (empty)"
"Headers : (4)"
Accept-Encoding : compress;q=0.5, gzip;q=1.0
Content-Type : application/x-www-form-urlencoded; application/json
Accept : application/json
Accept-API-Version : resource=2.0
```

content type now for whatever reason is **Content-Type : application/x-www-form-urlencoded; application/json**

The server in this instance cant handle that, but the issue is that Fuel is doing something other than what the code tells it to do, it should send **Content-Type : application/json** as that is what was specified in code

### LINKED ISSUE #413: Content-Type application/x-www-form-urlencoded by default
Hi all!

I updated from 1.14.0 to 1.15.0 and had some issues with all my requests getting broken.
Before I was just doing very basic JSON PUT requests, like this
```
val (request, response, result) = "http://myserver.local/api/v1/test".httpPut()
                .header("Content-Type" to "application/json")
                .body("{"id":"test","somemorefields":"andvalues", ...")
                .response()
```

with the latest Update to 1.15, this doesn't work anymore. The Content-Type being sent is then "application/x-www-form-urlencoded; application/json" and my server rejects this.

Is this a nice default behaviour? I wouldn't expect a default header to be set and my explicitly defined headers sent on the second position only (which makes everything fail now). I guess this whole implementation changed with https://github.com/kittinunf/Fuel/pull/393...

Sidenote for anyone with the same problem: I figured out that the headers can be reset by wrapping them in a map 
```
... .header(mapOf("Content-Type" to "application/json"))...
```
Kind Regards
Matthias

### LINKED ISSUE #472: jsonBody function leads to two different content-type headers in post request
# Bug Report

## Description
Using Fuel 1.16.0 and the json data method introduced in #418 in a post request leads to two different content-type headers:
* `Content-Type=application/x-www-form-urlencoded`
* `content-type=application/json`

## To Reproduce

```Kotlin
println(Fuel.post("http://localhost:8080").jsonBody("{}").headers)
```
Output:
```
{Accept-Encoding=compress;q=0.5, gzip;q=1.0, Content-Type=application/x-www-form-urlencoded, content-type=application/json}
```

## Expected behavior
There is only one header `Content-Type=application/json`.

### LINKED ISSUE #473: jsonBody does not replace default content encoding

### LINKED ISSUE #478: Headers fields must be case insensitive
# Bug Report

## Description
From the spec:

> Each header field consists of a case-insensitive field name followed
   by a colon (":"), optional leading whitespace, the field value, and
   optional trailing whitespace.

     header-field   = field-name ":" OWS field-value OWS

     field-name     = token
     field-value    = *( field-content / obs-fold )
     field-content  = field-vchar [ 1*( SP / HTAB ) field-vchar ]
     field-vchar    = VCHAR / obs-text

     obs-fold       = CRLF 1*( SP / HTAB )
                    ; obsolete line folding
                    ; see Section 3.2.4

>   The field-name token labels the corresponding field-value as having
   the semantics defined by that header field.  For example, the Date
   header field is defined in Section 7.1.1.2 of [RFC7231] as containing
   the origination timestamp for the message in which it appears.

## To Reproduce

1. add a header: `"Cache-Control": "no-cache"`
2. add the same header with different casing: "cache-control": "max-age: 600"
3. inspect the request and see the header twice.

## Expected behavior
The last assignment survives _or_ if the method allows for appending, the headers are combined.

## Additional context
https://tools.ietf.org/html/rfc7230#section-3.2

Leads to bugs such as #474 #472 #413

SleeplessByte:
## Description

This PR tackles a lot of RFC violations, most notably making the headers normalised on both `Request` and `Response`. This means that `Content-Type` is the same header as `content-type`, as defined by the RFC. Furthermore it fixes the special behaviour for `Cookie`, which should be joined by a `;` as per the RFC, and the other **listable** headers by a `,`. Some headers can always only be a single value, and that is also enforced.

Some of these were already closed with a temporary workaround, but this:

Fixes #478 
Fixes #474
Fixes #473
Fixes #472 
Fixes #413 
Fixes #408 / Reverts "API"
Fixes #392

It is important to understand that this:
- reverts the behaviour of `header` being overwrite by default. Overwriting a header is far more common than appending. The behaviour is supported by a new function `appendHeader` which does exactly what it says.
- completely stops `baseHeaders` from being applied, if the request has them. We might want to change this to always ADD the `request headers to the base headers. I think the replace approach is more intuitive.
- adds accessors (`[header]=` and `[header]`) to request and a getter `[header]` to response

What can be done further?
- Cookies returned via `Set-Cookie` should be handled by a CookieJar #263 
- Parameters should probably get their own structure, easy to convert to, as well.

## Type of change

Check all that apply

## How Has This Been Tested?

Tests have been changed where necessary, but *ALL* the other tests pass without change.

## Checklist:


iNoles:
It doesn't like Wildcard import :)

README.md
```diff
@@ -621,42 +621,54 @@ object Windows1255StringDeserializer : ResponseDeserializable<String> {
...
- `fun header(pairs: Map<String, Any>?): Request` takes a map this method can not support multiple
-  values for the same key as per the definition of the map.
+* `Headers` can be added to a request via various methods including
+    - `fun header(name: String, value: Any): Request`: `request.header("foo", "a")`
+    - `fun header(pairs: Map<String, Any>): Request`: `request.header(mapOf("foo" to "a"))`
+    - `fun header(vararg pairs: Pair<String, Any>): Request`: `request.header("foo" to "a")`
+    - `operator fun set(header: String, value: Collection<Any>): Request`: `request["foo"] = listOf("a", "b")`
+    - `operator fun set(header: String, value: Any): Request`: `request["foo"] = "a"`
+* By default, all subsequent calls overwrite earlier calls, but you may use the `appendHeader` variant to append values to existing values.
+    - In earlier versions a `mapOf` overwrote, and `varargs pair` did not, but this was confusing.
```
markGilchrist: have you added a Jdoc comment on the methods that have changed so that people who use the method know the behaviour has changed?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
```diff
@@ -0,0 +1,225 @@
...
+        fun isSingleValue(key: HeaderName): Boolean {
+            return when (key) {
+                // These headers, per RFC, SHOULD only appear once.
+                HeaderName(AGE) -> true
+                HeaderName(CONTENT_TYPE) -> true
+                HeaderName(CONTENT_LENGTH) -> true
+                HeaderName(CONTENT_LOCATION) -> true
+                HeaderName(EXPECT) -> true
+                HeaderName(EXPIRES) -> true
+                HeaderName(LOCATION) -> true
```
markGilchrist: would you consider having a collection of header names called `singleValueHearders` and chaining this to

`return singleValueHeaders.contains(key)`? 

The above collection feels like a magic variable and by naming the collection it explains what makes a single value.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
```diff
@@ -34,15 +37,6 @@ class Request(
     var timeoutInMillisecond: Int,
     var timeoutReadInMillisecond: Int
 ) : Fuel.RequestConvertible {
-
-    @Deprecated(replaceWith = ReplaceWith("method"), message = "http naming is deprecated, use 'method' instead")
-    val httpMethod
-        get() = method
-
-    @Deprecated(replaceWith = ReplaceWith("headers"), message = "http naming is deprecated, use 'headers' instead")
```
markGilchrist: should we be removing these now?

markGilchrist: @kittinunf

README.md
```diff
@@ -112,7 +112,7 @@ dependencies {
 #### Async mode
 
 * Kotlin
-``` Kotlin
```
kittinunf: [Q] Does it make any difference?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    override fun toString(): String {
+        return name
+    }
+}
+
+typealias HeaderValues = Collection<String>
+
+class Headers : MutableMap<String, HeaderValues> {
```
kittinunf: Since most of the implementation is passing through to underlying hashMap contents, shall we use constructor delegation?

``` Kotlin
class Header(private val contents: HashMap<HeaderName, HeaderValues>) : MutableMap<HeaderName, HeaderValues> by contents {
  // only override something that is custom for example remove or some other
}
```

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
kittinunf: Since this is a breaking change, let's do it 🎉.

Perhaps, this will be a major bump for the next version, do you 2 think it makes sense?

README.md
```diff
@@ -621,42 +621,54 @@ object Windows1255StringDeserializer : ResponseDeserializable<String> {
...
+* By default, all subsequent calls overwrite earlier calls, but you may use the `appendHeader` variant to append values to existing values.
+    - In earlier versions a `mapOf` overwrote, and `varargs pair` did not, but this was confusing. 
 
-``` Kotlin
-Fuel.get("/get").response { request, response, result ->
-    //make request to https://httpbin.org/get with global device header (Device : Android)
-}
+```kotlin
+Fuel.post("/my-post-path")
+    .header("Accept", "text/html, */*; q=0.1")
```
kittinunf: Shall we encourage user to use our internal `const val` too? like `CONTENT_TYPE`. if not so we'd better mark them as private?

README.md
SleeplessByte: Yes, they are documented. I can add a note that the behaviour changed.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
SleeplessByte: @kittinunf I want the headers to be accessible by any `String` and not force users to wrap in `HeaderName`, so unfortunately that won't work.

Down the line, we probably want `Fuel` to be more immutable (and thus threadsafe) and change this to `Map<HeaderName, HeaderValues>` anyway, which will require MORE custom implementation.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
SleeplessByte: It would be a constant then: `SINGLE_VALUE_HEADERS = listOf(...)` which has a `O(n)`. I don't care for either. This just organically grew.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt
SleeplessByte: Yes!

README.md
SleeplessByte: Yep. Not on GitHub, but it does on compiling jdoc/kdoc, and various plugins in Intelij/Android Studio.

README.md
SleeplessByte: Added in Kdoc!

README.md
SleeplessByte: Changed to use our constants!

SleeplessByte:
> Do you feel that some more unit tests are needed?
@markGilchrist defo.

I first wanted to get everyone's thoughts on this as it is quite a big change. Don't want to overstep anyone or anything, but working to a more RFC compliant Fuel. If this PR is appreciated and approved, it opens the door for my streams input and output PR which solves another array of issues.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
kittinunf: That's unfortunate ...

kittinunf: Come to think about this by `users`, do you mean users who use Request class or the Request class itself? If it is the former, we can actually still expose only the API that interacts with `String` but internally (in Request class) we can do the wrapping in implementation? I would say if the wrapping `HeaderName` part is internal, it should be _fine_?

like

``` Kotlin 
operator fun get(header: String): HeaderValues {
  return headers[HeaderName(header)]
}

operator fun set(header: String, values: Collection<*>): Request {
  headers[HeaderName(header)] = values.map { it.toString() }
  return this
}
```

so on and so forth.

SleeplessByte: I meant both: I see people do `request.headers.things`. 

I don't mind changing it tho. I do like that the logic of normalisation is actually in `Headers` and not in both `Request` and `Response`. The downside of your proposal is having that on both ends, which could be fine.

kittinunf: Yes you are right. I guess the benefit of my proposal is we have to only maintain the code that only matter. 

I am not super against leaving it like this too though. I would say I'll leave the final judgement to you.

SleeplessByte: I think we leave it like this for now, it's not too much work to change later -- which we'll probably do anyway to make Fuel completely threadsafe (read: immutables)

SleeplessByte:
Tests are coming!

fuel/src/test/kotlin/com/github/kittinunf/fuel/core/HeadersTest.kt
```diff
@@ -0,0 +1,219 @@
...
+
+        val contentTypeSets = recordSet[Headers.CONTENT_TYPE]
+        val expectedContentType = headers[Headers.CONTENT_TYPE].last()
+
+        assertThat("Expected CONTENT_TYPE to be added via 'set'", contentTypeSets, notNullValue())
+        assertThat(contentTypeSets, isEqualTo(expectedContentType))
+
+        assertThat(recordSet.size, isEqualTo(2))
+        assertThat(recordAdd.size, isEqualTo(1))
+
```
kittinunf: **[ktlint]** <sub>reported by [reviewdog](https://github.com/haya14busa/reviewdog) :dog:</sub>
Unexpected blank line(s) before "}"

SleeplessByte:
🏃 tests and documentation added, please review and perhaps merge!

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
markGilchrist: the above implementation is O(n) the only thing that would change would be the syntax

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
```diff
@@ -49,44 +63,52 @@ class Headers : MutableMap<String, HeaderValues> {
...
+     * Appends a single value [value] to the current values at [header]
+     *
+     * @param header [String] the header to append to
+     * @param value [Any] the value to append, coerced with [Any.toString]
+     */
+    fun append(header: String, value: Any): Headers {
+        return when (Headers.isSingleValue(header)) {
+            true -> set(header, value.toString())
+            false -> set(header, this[header].plus(value.toString()))
         }
```
markGilchrist: would you consider 

``` kotlin
fun append(header: String, value: Any) = set(header, if(Headers.isSingleValue(header)value.toString() else this[header].plus(value.toString())))
        
```

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
```diff
@@ -116,13 +138,31 @@ class Headers : MutableMap<String, HeaderValues> {
...
+     * @return [HeaderValues] the values at [key] or an empty list
      */
     override fun get(key: String): HeaderValues {
-        return contents[HeaderName(key)].orEmpty()
+        val header = HeaderName(key)
+        return contents[header].orEmpty().let {
+            when (Headers.isSingleValue(header)) {
+                true -> listOf(it.last())
+                false -> it
+            }
```
markGilchrist: could you consider 
``` kotlin

override fun get(key: String) = contents[HeaderName(key)].orEmpty().let {
    if (Headers.isSingleValue(header)) listOf(it.last()) else it
}
```

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
```diff
@@ -180,37 +264,67 @@ class Headers : MutableMap<String, HeaderValues> {
...
+         *  A recipient MAY combine multiple header fields with the same field
+         *  name into one "field-name: field-value" pair, without changing the
+         *  semantics of the message, by appending each subsequent field value to
+         *  the combined field value in order, separated by a comma.  The order
+         *  in which header fields with the same field name are received is
+         *  therefore significant to the interpretation of the combined field
+         *  value; a proxy MUST NOT change the order of these field values when
+         *  forwarding a message.
+         *
+         * @param header [HeaderName] the header to check
```
markGilchrist: awesome explanations!

SleeplessByte:
Yes sure. I'll make it a map with default false. That way it's O(1)!

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
kittinunf: I don't see this list is getting anywhere near something that o(n) can be significantly slow though. Just aim for better readability.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
SleeplessByte: To me that looks horrible, but if that's our coding style, sure!

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
SleeplessByte: To me that looks horrible, but if that's our coding style, sure!

kittinunf:
``` Kotlin
fun append(header: String, value: Any): Headers {
  val value = if(Headers.isSingleValue(header)) value.toString() else this[header].plus(value.toString())
  return set(header, value)
}
```

kittinunf:
Oh the test fails though. Lemme check.

kittinunf:
```
com.github.[secure].fuel.RequestHeadersTest > requestHeadersFromReadmeIntegrity FAILED
    java.lang.AssertionError at RequestHeadersTest.kt:29
com.github.[secure].fuel.RequestHandlerTest > httpGetRequestWithMalformedHeaders FAILED
    junit.framework.AssertionFailedError at RequestHandlerTest.kt:67
com.github.[secure].fuel.RequestAuthenticationTest > httpBasicAuthenticationWithValidCase FAILED
    java.lang.AssertionError at RequestAuthenticationTest.kt:82
```

Something to do with your last commit?

SleeplessByte:
Great!

I would get this merged and then focus on the 2 subsequent body PRs which
basically close half the outstanding tickets

SleeplessByte:
I personally like the when syntax better visually. I kinda miss my ternary
:(

PR #485, opened 2018-10-26
Better bodies and encoding

SleeplessByte:
## Description

This PR overhauls the way bodies are built and written to the stream, as well as various additions and fixes to bodies, upload bodies and the default HttpClient.

> I'm opening this so everyone can see where this is going and comment on it.
> *Still working on that one failing test which I haven't figured out why*.

- Should allow #464 to be written by wrapping a `Body` by a `Body` which stores the entire output into an intermediary / memory, calculating the signature, and moving on.
- From hereon out the streams are buffered so #146 could be implemented on top of this
- This is prework to fix the datapart/blob issues

This also:
- Remove the need to call the body callback twice
- Add closing the input streams when they are written to the body
- Add buffering the output stream correctly, so it's not all loaded in memory
- Add buffering the input stream correctly, so it's not all loaded in memory
- Add chunked encoding for unknown / dynamic file sizes, which are then efficiently pumped through the connection
- Add Content-Encoding ánd Transfer-Encoding handling, based on the HttpURLConnection docs
- Add support for `deflate` encoding
- Add HTTP Cache by default, with an opt-out on manager or request basis
- Add opt-out on manager or request basis for content decoding

## Type of change

Check all that apply

- [X] Bug fix (a non-breaking change which fixes an issue)
- [X] New feature (a non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] Refactoring (a change which changes the current internal or external interface)
- [X] This change requires a documentation update

## How Has This Been Tested?

In case you did not include tests describe why you and how you have verified the
changes, with instructions so we can reproduce. If you have added comprehensive
tests for your changes, you may omit this section.

## Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation, if necessary
- [ ] My changes generate no new compiler warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Inspect the bytecode viewer, including reasoning why

Issue #479, opened 2018-10-22
Improve multipart features for httpPost and httpUpload


abelsromero:
# Feature Request

With the current API, some scenarios of multipart interaction are not friendly and require what imho are hacks.

## Description
I am working with CMS API and there's a API call with requires uploading a document alongside metadata in JSON.
This is easily done in curl with:
```cUrl
curl "https://api.hostname/v1/documents" \
-X POST \
-H 'Content-Type: multipart/form-data' \
-H "Authorization: Bearer $token" \
-H "Repository-Id: $repo_id" \
-F metadata="{
    \"description\": \"Test document\",
    \"categories\": [{...}]
	};type=application/json" \
-F file=@file_path
```

I only got this to work with the following code

```kotlin

    val (request, response, result) = url.httpUpload(
//            parameters = listOf("metadata" to """{"description":"test document from Kotlin"};type=application/json""")
    )
            .dataParts { request, url ->
                listOf(
                        DataPart(File("src/main/resources/rust-logo-blk.svg"), "file"),
                        DataPart(File("src/main/resources/doc.json"), "metadata","application/json")
                )
            }
            .header("Authorization" to "Bearer $token")
            .header("Repository-Id" to repoId)
            .responseString()
```

As you can see this requires storing the JSON in a file. That's because I could not find any way to set the Content-Type for the part other than using `DataPart`. Using `parameters` as seen in the commented line sets it as `text/plain` causing the service to return an error.

I also tried something like

```kotlin
    val (request, response, result) = url.httpPost(listOf(
            "metadata" to """{"description":"test document from Kotlin"}""",
            "file" to DataPart(File("resources/rust-logo-blk.svg"))))
            .header("Content-Type" to "multipart/form-data")
            .header("Authorization" to "Bearer $token")
            .header("Repo-Id" to repoId)
            .responseString()
```

## Proposed Solution
Imho, the naming `upload` is confusing for this cases but is very useful for simple 1 file uploads. I'd advocate to:
1. Create a httpMultipart similar to httpPost that offers the ability to pass any kind of object, not only files and set the Content-Type. Or allow htttPost to set Content-Type to multipart so that it manages the format internally (a la curl).
2. Extend DataPart to support any data (String, Inputstream, ...)

Probably, with only point 2 would be enough to support a cleaner method with httpUpload.

## Alternatives I've considered
Current solution does not allow to generate metadata Json on-the fly, which is not valid for a dynamic system.
[...]

Issue #464, opened 2018-10-17
How to get requestbody inside interceptor


jonni-larjomaa:
I'm writing an interceptor which makes an authorization header which is calculated over the entire request payload, i need to have the body of the request in this calculation also but there is no obvious way to get the request body ?

This calculation is needed for example AWS v4 request signing.

Solution: 
Request obejcts getHttpBody should be public ?

SleeplessByte:
Hello @jonni-larjomaa, 

https://github.com/kittinunf/Fuel/pull/485/files#diff-872bb60dbe62daac0dfbc7740a3379a9 should cover you for now. I do want to support AWS signing out of the box at some point, so if you create an implementation, we would 💘 a PR.

Issue #463, opened 2018-10-16
Add ability to customize DownloadTaskReqeust FileOutputStream


charshep:
# Feature Request

## Description
I'd like to incorporate [RNCrytpor-Native](https://github.com/TGIO/RNCryptorNative) in conjunction with Fuel in order to encrypt downloaded files. Unencrypted data should never be written to storage, even temporarily, so I'd like to wrap/override the `FileOutputStream` that is used internally by `DownloadTaskRequest`.

## Proposed Solution
I'm not familiar enough yet with Fuel to propose an exact design but the ideas discussed in #225 seem like they would lend themselves to accommodating configuring streams used by Fuel.

## Alternatives I've considered
`ResponseDeserializable`, if I understand it correctly, isn't appropriate because it would only let me rewrite a downloaded file that has already been written to disk without encryption.

charshep:
I'm starting to work on a fork of this. I'm having trouble including `rncryptor-native` however. In a separate application project I'm able to simply add:

    implementation 'com.github.tgio:rncryptor-native:0.0.9'

to my build.gradle and I can import it no problem. However, adding the same to `Fuel/fuel/build.gradle` while it does in fact pull down the library (I can see it in External Libraries in Project view) ,  `import` statements in code fail to properly import the file and the usual auto-import alt-return keyboard shortcut can't seem to find the library.

Anyone able to help me out with what I'm missing here? Sorry if this is obvious, still pretty new to Kotlin. Apologies also if there are more appropriate place to post this kind of question.

SleeplessByte:
The IDE can definitely fuck this up.

What I've tried, if invalided and restart doesn't work, is actually
deleting the .idea folder and build again from gradle.

iNoles:
It looks like that only work in Android.

charshep:
Thanks @SleeplessByte

@iNoles Yep, realized that after a bit. Thanks. :P

SleeplessByte:
I'm working on an implementation that allows for generic streams (#479), that should also cover this.

PR #486, opened 2018-10-28
Generic Multipart bodies, upload and download progress

### LINKED ISSUE #193: Upload progress is actually progress of creating a temp file?
In [UploadTaskRequest.kt](https://github.com/kittinunf/Fuel/blob/master/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/UploadTaskRequest.kt), the progress callback isn't sending the progress of the file upload, it's incorrectly sending the progress of writing the file to the output stream

Line 42:
```
    //input file data
    if (outputStream != null) {
       FileInputStream(file).use {
           it.copyTo(outputStream, BUFFER_SIZE) { writtenBytes ->
               progressCallback?.invoke(contentLength + writtenBytes, totalLength)
             }
         }
    }
  contentLength += file.length()
  contentLength += write(CRLF)
```

So does that mean when progress is 100%,  that the upload has been submitted to the server, but it doesn't mean that the upload has completed uploading to the remote server?

### LINKED ISSUE #344: How to upload customed Blobs (like DataPart) in multipart way?
I have a Blob list to upload with different names and filenames like

`listOf(blob("a","a1.jpg",size,stream),blob("a","a2.jpg",size,stream),
blob("b","b1.jpg",size,stream),blob("b",b2.jpg",size,stream))`

are there any way to do it with fuel like DataParts?

`DataPart(Blob(),"name", "image/jpeg")`

 or is this right for this situation

```kotlin
fun compareApi(faceImgData: FaceImgData,callback: (Result<CompareCBData, FuelError>)->Unit){
        compare_suffix.httpUpload(Method.POST).blob { request, url ->
            faceImgData.fullImg
        }.name {
            "fullImg"
        }.blobs { request, url ->
            faceImgData.faces
        }.name {
            "faces"
        }.responseObject(CompareCBData.Deserializer()) { request, response, result ->
            callback(result)
        }
    }
```

### LINKED ISSUE #399: multipart/form-data boundary isn't set correctly
```
val result = FuelManager()
  .request(Method.PUT, "/", listOf("lat" to "3"))
  .apply { type = Request.Type.UPLOAD }
  .dataParts { request, url ->
    listOf(DataPart(File("some.jpg")))
  }.awaitStringResult()
```
This made some trouble with ktor server.
```
java.io.IOException: Failed to parse multipart: Content-Type should be multipart/* but it is application/x-www-form-urlencoded
```
[`Request.dataParts`](https://github.com/kittinunf/Fuel/blob/master/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Request.kt#L182) doesn't seem set boundary header, and [`UploadTaskRequest.bodyCallBack`](https://github.com/kittinunf/Fuel/blob/master/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/UploadTaskRequest.kt#L18) uses arbitrary boundary in this situation. It leads to
```
PUT http://localhost/ HTTP/1.1
Accept-Encoding: compress;q=0.5, gzip;q=1.0
Content-Type: application/x-www-form-urlencoded
Cache-Control: no-cache
Pragma: no-cache
User-Agent: Java/1.8.0_172
Host: localhost:8080
Accept: text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2
Connection: keep-alive
Content-Length: 209263

--164e202956e
Content-Disposition: form-data; name="lat"
Content-Type: text/plain
```
Content-Type is urlencoded, but body is multipart/form-data, and library user should set boundary manually. Is it intended?

### LINKED ISSUE #416: Fuel.download doesn't handle redirected url correctly
```kotlin
    Fuel.download("https://github.com/kittinunf/Fuel/releases/download/1.12.1/fuel-1.12.1.jar")
            .destination { response, url ->
                File("fuel.jar")
            }.progress { readBytes, totalBytes ->
                val progress = readBytes.toFloat() / totalBytes.toFloat()
                println(progress)
            }.response { req, res, result ->
            }
```

The `redirectResponseInterceptor` added to Fuel.download call a new request `next(request, manager.request(encoding).header(newHeaders).response().second)`. Thus `destination` and `progress` won't work as they are bound to the first request.

SleeplessByte:
> Builds on #485, this time tacking everything `multipart`, meaning `Blob`, `DataPart` and so forth.

> - Extracts request client options to the executor, instead of having them live on request. They only make sense conditionally and are not actually part of the request themselves. We keep growing the list of attributes, and this solves that.
> - Composition for Download and Upload by wrapping request and overriding "this" (by using the "request" property). This allows us to conditionally make functions available and not rely on an enumeration. This also opens up a request that uploads AND downloads.
> - Request cancelability was actually not implemented correctly. Some requests are currently still not cancellable! This change actually only exposes .cancel if the request is.

**EDIT**: Checkboxes indicate which have been extracted out into new PRs.

- [ ] Fixes #399
- [ ] Fixes #344
- [ ] Should allow to implement #463 with an interceptor
- [X] Should allow to implement #464 with an interceptor
- [ ] Adds unknown lengths (blob, datapart)
- [X] Adds generic streams for data parts #479 
- [ ] Add naming blobs via `DataPart.from(blob)` #344 
- [X] Add `request` / `upload` progress for *all* requests #146 
- [X] Fix progress from an original request to a redirected request #416
- [X] Fix progress using temporary byte array #193


fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/RedirectionInterceptor.kt
```diff
@@ -29,26 +29,29 @@ fun redirectResponseInterceptor(manager: FuelManager) =
...
-                // check whether it is the same host or not
-                if (newUrl.host != request.url.host) {
-                    newHeaders.remove(Headers.AUTHORIZATION)
-                }
+                val encodedRequest = Encoding(
+                    httpMethod = when {
+                        response.statusCode in redirectStatusWithGets -> Method.GET
+                        else -> request.method
+                    },
+                    urlString = when(URI(redirectedUrl).isAbsolute) {
```
kittinunf: **[ktlint]** <sub>reported by [reviewdog](https://github.com/haya14busa/reviewdog) :dog:</sub>
Missing spacing after "when"

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/RedirectionInterceptor.kt
```diff
@@ -29,26 +29,29 @@ fun redirectResponseInterceptor(manager: FuelManager) =
...
+                        true -> URL(redirectedUrl).toString()
+                        false -> URL(request.url, redirectedUrl).toString()
+                    }
+                ).request
+
+                val newRequest = manager.request(encodedRequest)
+                    .header(
+                        Headers.from(request.headers).let {
+                            when(encodedRequest.url.host != request.url.host) {
```
kittinunf: **[ktlint]** <sub>reported by [reviewdog](https://github.com/haya14busa/reviewdog) :dog:</sub>
Missing spacing after "when"

fuel-coroutines/src/main/kotlin/com/github/kittinunf/fuel/coroutines/Coroutines.kt
```diff
@@ -118,11 +118,3 @@ suspend fun <U : Any> Request.awaitObjectResult(
     }
     Result.Failure(fuelError)
 }
-
-@Deprecated("please use 'awaitByteArray()'", ReplaceWith("awaitByteArray()", "deserializable"))
-suspend fun Request.awaitResponseResult(): ByteArray = awaitByteArray()
-
-@Deprecated("please use 'awaitObjectResult(deserializable)'", ReplaceWith("awaitObjectResult(deserializable)"))
-suspend fun <U : Any> Request.awaitSafelyObjectResult(
-    deserializable: ResponseDeserializable<U>
```
markGilchrist: do we have a policy on how long we leave deprecated methods in the lib before we remove them?

fuel-rxjava/src/main/kotlin/com/github/kittinunf/fuel/rx/RxFuel.kt
```diff
@@ -11,25 +11,25 @@ import com.github.kittinunf.result.Result
 import io.reactivex.Single
 import java.nio.charset.Charset
 
-fun Request.rx_response() = rx_response(ByteArrayDeserializer())
+fun Request.toRxResponse() = toRxResponse(ByteArrayDeserializer())
 
-fun Request.rx_responseString(charset: Charset = Charsets.UTF_8) = rx_response(StringDeserializer(charset))
+fun Request.toRxResponseString(charset: Charset = Charsets.UTF_8) = toRxResponse(StringDeserializer(charset))
```
markGilchrist: great

fuel/src/main/kotlin/com/github/kittinunf/fuel/Fuel.kt
```diff
@@ -30,89 +33,89 @@ class Fuel {
         // get
         @JvmStatic
         @JvmOverloads
-        fun get(path: String, parameters: List<Pair<String, Any?>>? = null): Request =
+        fun get(path: String, parameters: Parameters? = null): Request =
```
markGilchrist: I think this is a great idea

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Blob.kt
```diff
@@ -3,7 +3,7 @@ package com.github.kittinunf.fuel.core
 import java.io.InputStream
 
 data class Blob(
-    val name: String = "",
+    val name: String,
```
markGilchrist: out of interest why did you remove this default arg?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Body.kt
```diff
@@ -38,28 +37,30 @@ data class DefaultBody(
...
-            openReader = CONSUMED_READER
+            openStream = CONSUMED_READER
+
+            length
         }
     }
 
-    override fun isEmpty() = openReader == null || (length != null && length == 0)
-    override fun isConsumed() = openReader === CONSUMED_READER
+    override fun isEmpty() = openStream == null || (length != null && length == 0)
```
markGilchrist: this could be
`override fun isEmpty() = openStream == null || length == 0`

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
```diff
@@ -32,26 +29,22 @@ class Encoding(
...
-        ).apply {
-            if (data != null) body(data)
+        RegularRequest(
+            method = method,
+            path = modifiedPath,
+            url = createUrl(modifiedPath),
+            parameters = parameters ?: emptyList(),
+            headers = headerPairs
+        ).let {
+            when (data) {
```
markGilchrist: would an `if` be more readable?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/MultipartBody.kt
```diff
@@ -0,0 +1,165 @@
...
+import java.io.ByteArrayInputStream
+import java.io.ByteArrayOutputStream
+import java.io.OutputStream
+import java.net.URLConnection
+import java.nio.charset.Charset
+
+data class MultipartBody(private val request: MultipartRequest) : Body {
+    private var inputAvailable: Boolean = true
+
+    override fun isConsumed() = !inputAvailable
```
markGilchrist: could this be a val with a private set?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/RegularRequest.kt
```diff
@@ -0,0 +1,407 @@
...
+     *  that would lead to closed streams being unable to be read. If the size is known, set it before anything else.
+     *
+     * @param openStream [BodySource] a function that yields a stream
+     * @param calculateLength [Number?] size in +bytes+ if it is known
+     * @param charset [Charset] the charset to write with
+     *
+     * @return [Request] the request
+     */
+    override fun body(openStream: BodySource, calculateLength: BodyLength?, charset: Charset): Request {
+        this.body = DefaultBody.from(openStream = openStream, calculateLength = calculateLength, charset = charset)
```
markGilchrist: great kDocs!

fuel-coroutines/src/main/kotlin/com/github/kittinunf/fuel/coroutines/Coroutines.kt
SleeplessByte: I tend to keep them around at least 2 major versions (so since 1.14), but just take the opportunity to remove them as we don't test them anymore

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/MultipartBody.kt
SleeplessByte: `isConsumed` always reflects `inputAvailable`. If it's a val, it will never change, right?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Encoding.kt
SleeplessByte: Personally I like the explicit return. I think this should actually be wrapped in a function `applyBody`, then we can use an if.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Blob.kt
SleeplessByte: Haven't figured out completely how I think this API should work, but mostly because `name` doesn't exist on request anymore. Explicit naming works much better. There are already a few tests generating `_0` as name, which does not make any sense, to me.

Removing this default, but allowing some sort of "default options" via the request and injecting it via `blob` would work.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/MultipartBody.kt
markGilchrist: sorry I meant 
``` kotlin
var isConsumed :Boolean 
    private set
```

SleeplessByte:
Keeping this open for now for tracking purposes. Please don't ❌ close.

This 🙅 won't be applied as a whole at once, but instead be broken up into multiple PRs 🗡

fuel-rxjava/src/main/kotlin/com/github/kittinunf/fuel/rx/RxFuel.kt
```diff
@@ -11,25 +11,25 @@ import com.github.kittinunf.result.Result
...
 
-fun <T : Any> Request.rx_responseObject(deserializable: Deserializable<T>) = rx_response(deserializable)
+fun <T : Any> Request.toRxResponseObject(deserializable: Deserializable<T>) = toRxResponse(deserializable)
```
kittinunf: Hmm I thought we have changed to `rxResponseObject` already in master which is good. Why do we change again here with prefixing it with `to`?

fuel-rxjava/src/main/kotlin/com/github/kittinunf/fuel/rx/RxFuel.kt
```diff
@@ -11,25 +11,25 @@ import com.github.kittinunf.result.Result
...
 
-private fun <T : Any> Request.rx_response(deserializable: Deserializable<T>): Single<Pair<Response, Result<T, FuelError>>> =
+private fun <T : Any> Request.toRxResponse(deserializable: Deserializable<T>): Single<Pair<Response, Result<T, FuelError>>> =
         rx {
             val (_, response, result) = response(deserializable)
             response to result
         }
 
-fun Request.rx_bytes() = rx_result(ByteArrayDeserializer())
+fun Request.toRxBytes() = toRxResult(ByteArrayDeserializer())
```
kittinunf: Here too.

kittinunf:
ah just realize that you need to rebase master first I think. So the changes make sense again?

iNoles:
Wow a lot of Conflicts.

PR #497, opened 2018-11-02
Add option to pass in a Serializer to kotlinxSerializerOf without propagating @ImplicitReflectionSerializer

NikkyAI:
## Description

this avoids infecting user code with `@ImplicitReflectionSerializer` when a DeserializationStrategy is passed by the user

annoyingly the annotation requires to create extra functions

## Type of change

Check all that apply

## How Has This Been Tested?

## Checklist:

- [ ] I have performed a self-review of my own code

### DIFF
--- a/fuel-kotlinx-serialization/src/main/kotlin/com/github/kittinunf/fuel/serialization/FuelSerialization.kt
+++ b/fuel-kotlinx-serialization/src/main/kotlin/com/github/kittinunf/fuel/serialization/FuelSerialization.kt
@@ -14,28 +14,44 @@ import kotlinx.serialization.ImplicitReflectionSerializer
 import kotlinx.serialization.json.JSON
 import kotlinx.serialization.serializer
 
-@ImplicitReflectionSerializer
 inline fun <reified T : Any> Request.responseObject(
-    loader: DeserializationStrategy<T> = T::class.serializer(),
+    loader: DeserializationStrategy<T>,
     json: JSON = JSON.plain,
     noinline deserializer: (Request, Response, Result<T, FuelError>) -> Unit
 ) = response(kotlinxDeserializerOf(loader, json), deserializer)
 
 @ImplicitReflectionSerializer
 inline fun <reified T : Any> Request.responseObject(
+    json: JSON = JSON.plain,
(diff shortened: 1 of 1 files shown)


fuel-kotlinx-serialization/src/main/kotlin/com/github/kittinunf/fuel/serialization/FuelSerialization.kt
```diff
@@ -14,28 +14,44 @@ import kotlinx.serialization.ImplicitReflectionSerializer
 import kotlinx.serialization.json.JSON
 import kotlinx.serialization.serializer
 
-@ImplicitReflectionSerializer
 inline fun <reified T : Any> Request.responseObject(
-    loader: DeserializationStrategy<T> = T::class.serializer(),
+    loader: DeserializationStrategy<T>,
```
kittinunf: It looks like you are removing default argument here ... and require a client to passing by himself/herself. Is it what you are expecting to do in this PR?

NikkyAI: i do this so that `@ImplicitReflectionSerializer` is only required where it is actually necessary
`T::class.serializer()` requires to set this
and users of the library will then also have to annotate their functions

but you can get a serializer and pass it in without using any reflection

for that we need a second function that is not annotated
since the annotation cannot be put on the default argument

another option is to implicitly hide that annotation
it is possible but i think it would be bad since it is hiding potential issues from the user

kittinunf: Ahh ... hmm, actually this `@ ImplicitReflectionSerializer` is a bit more annoying than I was thinking originally.

What would be the client user code actually? Should we add something to show them in our README?

NikkyAI: the code does not change and idea will give users the quickfix to annotate it, i think it is not required

PR #502, opened 2018-11-04
Remove SameThreadExecutorService

iNoles:
## Description

SameThreadExecutorService doesn't use anywhere in the code. 

## Type of change

Check all that apply

## How Has This Been Tested?

## Checklist:


SleeplessByte:
Don't we use this in testing anymore?

iNoles:
@SleeplessByte  Nope, it is only when blocking is set to True. By default, it is always False.

kittinunf:
NICE

Issue #522, opened 2018-11-10
The sample app is crashing in current master


worker8:
# Bug Report
The sample app is crashing in current master

## Description

This is the crash log:
```
--------- beginning of crash
2018-11-10 14:38:59.652 30783-30783/com.example.fuel E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.example.fuel, PID: 30783
    java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1374)
     Caused by: java.lang.reflect.InvocationTargetException
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:327)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1374) 
     Caused by: java.lang.IllegalStateException: The input has already been written to an output stream and can not be consumed again.
        at com.github.kittinunf.fuel.core.DefaultBody$Companion$CONSUMED_STREAM$1.invoke(Body.kt:122)
        at com.github.kittinunf.fuel.core.DefaultBody$Companion$CONSUMED_STREAM$1.invoke(Body.kt:116)
        at com.github.kittinunf.fuel.core.DefaultBody.writeTo(Body.kt:94)
        at com.github.kittinunf.fuel.core.DefaultBody.toByteArray(Body.kt:78)
        at com.github.kittinunf.fuel.core.Request.cUrlString(Request.kt:509)
        at com.example.fuel.MainActivity$httpPut$1.invoke(MainActivity.kt:149)
        at com.example.fuel.MainActivity$httpPut$1.invoke(MainActivity.kt:33)
        at com.github.kittinunf.fuel.core.DeserializableKt$response$1.invoke(Deserializable.kt:49)
        at com.github.kittinunf.fuel.core.DeserializableKt$response$1.invoke(Unknown Source:4)
        at com.github.kittinunf.fuel.core.DeserializableKt$response$5$1.invoke(Deserializable.kt:76)
        at com.github.kittinunf.fuel.core.DeserializableKt$response$5$1.invoke(Unknown Source:0)
        at com.github.kittinunf.fuel.core.Request$callback$1.run(Request.kt:435)
        at android.os.Handler.handleCallback(Handler.java:789)
        at android.os.Handler.dispatchMessage(Handler.java:98)
        at android.os.Looper.loop(Looper.java:164)
[...]

Issue #493, opened 2018-11-01
Add support for Stetho Android debug bridge


igorwojda:
# Feature Request
Add Support for Stetho, so we can easily review all requests made by the Android application
http://facebook.github.io/stetho/

## Description
Currently reviewing underlying requests is quite hard (no support for Stetho). Besides that, this is very popular tools, so it's a shame that switching from Retrofit devs need to make the sacrifice.

## Proposed Solution
Here is some old implementation that I believe is a good place to start
https://github.com/Gira-X/fuel-stetho-client
however, it's a bit outdated, so it is not compatible with the newest version of Fuel

## Additional context
_Edit by @SleeplessByte_

From the documentation, we can see there are OkHttp interceptors available, so it will be easy to trivial to implement this for us:
> **Enable Network Inspection**
> If you are using the popular OkHttp library at the 2.2.x+ or 3.x release, you can use the Interceptors system to automatically hook into your existing stack. This is currently the simplest and most straightforward way to enable network inspection.
>
> **For OkHttp 2.x**
> ```Java
> OkHttpClient client = new OkHttpClient();
> client.networkInterceptors().add(new StethoInterceptor());
> ```
>
> **For OkHttp 3.x**
> ```Java
> new OkHttpClient.Builder()
>     .addNetworkInterceptor(new StethoInterceptor())
>     .build();
> ```
> 
> As interceptors can modify the request and response, add the Stetho interceptor after all others to get an accurate view of the network traffic.
>
> If you are using `HttpURLConnection`, you can use `StethoURLConnectionManager` to assist with integration though you should be aware that there are some caveats with this approach. In particular, you must explicitly add `Accept-Encoding: gzip` to the request headers and manually handle compressed responses in order for Stetho to report compressed payload sizes.
>
> See the stetho-sample project for more details.

iNoles:
Fuel uses HttpURLConnection.

SleeplessByte:
Yes, but that doesn't matter

iNoles:
This issue should be closed for #523

PR #523, opened 2018-11-11
Integrate Stetho

worker8:
## Description
This PR adds Stetho integration to HttpURLConnection.

## How to test this?
1. Launch `sample/MainActivity`
2. Open `chrome://inspect/#devices` in your Chrome browser
3. Click `inspect`:

| |
| - |
| |
4. Click `Go Couroutines` button.
- Note*: Due to the crash ([this issue](https://github.com/kittinunf/Fuel/issues/522)) in the sample app now, I cannot test different type of request other than `GET`.
5. Check the response:

image

## How it works
According to [this issue](https://github.com/facebook/stetho/issues/461#issuecomment-264560920), 

For Stetho to work, we need to use [`StethoURLConnectionManager`](https://github.com/facebook/stetho/blob/master/stetho-urlconnection/src/main/java/com/facebook/stetho/urlconnection/StethoURLConnectionManager.java), with this dependency: `implementation 'com.facebook.stetho:stetho-urlconnection:1.5.0'`.

There are 4 methods that we need to call inside [HttpClient.kt](https://github.com/kittinunf/Fuel/blob/master/fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt):

- StethoManager#preConnect(connection: HttpURLConnection, requestEntity: SimpleRequestEntity?)
- StethoManager#interpretResponseStream(inputStream: InputStream)
- StethoManager#postConnect()
- StethoManager#httpExchangeFailed(ioException: IOException)

`StethoManager` is an Android dependency, so we cannot use this inside `HttpClient.kt` directly, so I expose the correct place for intercepting through `stethoHook`. Then I make another module to actually make the call for calling the 4 methods.

This way, only users that want to use `Stetho` needs to include this module.

## How to use it (user perspective)
To use `Stetho` for `Fuel`, they need to use the following steps:
1. add `compile 'com.github.worker8.fuel:fuel-stetho:<latest-version>'`
 - not sure if I should to `com.github.kittinunf.fuel` to make it consistent?
2. add it to `FuelManager`:
```
FuelManager.instance.apply {
            stethoHook = StethoHook("Fuel Sample App")
        }
```
3. Done! 🎉 

TODO:
- [x] use `stetho.httpExchangeFailed()` in the correct place
- [x] make `stethoHook` to become generic hook, so that it can be in a separate module

## Type of change

Check all that apply

- [x] This change requires a documentation update

## How Has This Been Tested?

## Checklist:

  - not hard to understand area
  - will do so if PR is ok 
  - since this is stetho integration, a good way to check it is to run `sample/MainActivity.kt`, then check `Stetho` output using `Chrome`
  - how to do this? 😳 

Relevant issue: https://github.com/kittinunf/Fuel/issues/493


SleeplessByte:
Sweet! Will look at this tonight :D

build.gradle.kts
```diff
@@ -22,7 +22,7 @@ allprojects {
     }
 }
 
-val androidModules = listOf("fuel-android", "fuel-livedata")
+val androidModules = listOf("fuel-android", "fuel-livedata", "fuel-stetho")
```
iNoles: Can Stetho to be uses outside of Android?

fuel/build.gradle.kts
```diff
@@ -2,6 +2,8 @@ plugins { java }
 
 dependencies {
     compile(Result.dependency)
+    compile(Stetho.plugin)
```
SleeplessByte: I don't think [fuel-core] should be dependent on stetho at all.

fuel/build.gradle.kts
```diff
@@ -2,6 +2,8 @@ plugins { java }
...
+    compile(Stetho.StethoUrlConnection.plugin)
```
SleeplessByte: I don't think [fuel-core] should be dependent on stetho at all.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/FuelManager.kt
```diff
@@ -17,7 +17,8 @@ import javax.net.ssl.SSLSocketFactory
 import javax.net.ssl.TrustManagerFactory
 
 class FuelManager {
-    var client: Client by readWriteLazy { HttpClient(proxy) }
+    var client: Client by readWriteLazy { HttpClient(proxy, stethoHook) }
+    var stethoHook: HttpClient.StethoHook? = null
```
SleeplessByte: It seems that the stethoHook usage is isolated to `HttpClient` and can be injected after it has been created. I don't think we need this field at all.

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -27,10 +27,12 @@ import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 import kotlin.coroutines.suspendCoroutine
 
-internal class HttpClient(
+class HttpClient(
     private val proxy: Proxy? = null,
+    private var stethoHook: StethoHook? = null,
```
SleeplessByte: It's really not private, but this disallows setting/resetting it later. Make public instead.

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -49,7 +51,7 @@ internal class HttpClient(
             continuation.resume(doRequest(request))
         } catch (exception: Exception) {
             continuation.resumeWithException(exception as? FuelError
-                ?: FuelError(exception, ByteArray(0), Response(request.url)))
+                    ?: FuelError(exception, ByteArray(0), Response(request.url)))
```
SleeplessByte: Please use indentation of 4, not 8

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -65,8 +67,8 @@ internal class HttpClient(
             instanceFollowRedirects = false
 
             request.headers.transformIterate(
-                { key, values -> setRequestProperty(key, values) },
-                { key, value -> addRequestProperty(key, value) }
+                    { key, values -> setRequestProperty(key, values) },
```
SleeplessByte: Please use indentation of 4, not 8

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -65,8 +67,8 @@ internal class HttpClient(
...
+                    { key, value -> addRequestProperty(key, value) }
```
SleeplessByte: Please use indentation of 4, not 8

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -85,8 +87,8 @@ internal class HttpClient(
 
             // The TE (Accept Transfer Encoding) can only be one of these, should match decodeTransfer.
             setRequestProperty(
-                Headers.ACCEPT_TRANSFER_ENCODING,
-                Headers.collapse(HeaderName(Headers.ACCEPT_TRANSFER_ENCODING), SUPPORTED_DECODING)
+                    Headers.ACCEPT_TRANSFER_ENCODING,
```
SleeplessByte: Please use indentation of 4, not 8

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -85,8 +87,8 @@ internal class HttpClient(
...
+                    Headers.collapse(HeaderName(Headers.ACCEPT_TRANSFER_ENCODING), SUPPORTED_DECODING)
```
SleeplessByte: Please use indentation of 4, not 8

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -96,13 +98,15 @@ internal class HttpClient(
 
             setDoOutput(connection, request.method)
             setBodyIfDoOutput(connection, request)
+            stethoHook?.preConnect(connection, request)
```
SleeplessByte: I prefer `hooks.preConnect`, where `hooks` takes multiple hooks, not just one.

Are you sure THIS is where `HttpURLConnection actually connects? The body here has been sent already, so it seems that connection happens way earlier.

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -96,13 +98,15 @@ internal class HttpClient(
...
         }
 
         val headers = Headers.from(connection.headerFields)
         val transferEncoding = headers[Headers.TRANSFER_ENCODING].flatMap { it.split(',') }.map { it.trim() }
         val contentEncoding = headers[Headers.CONTENT_ENCODING].lastOrNull()
         var contentLength = headers[Headers.CONTENT_LENGTH].lastOrNull()?.toLong()
-        val shouldDecode = (request.decodeContent ?: decodeContent) && contentEncoding != null && contentEncoding != "identity"
+        val shouldDecode = (request.decodeContent ?: decodeContent) &&
+                contentEncoding != null && contentEncoding != "identity"
```
SleeplessByte: Please use indentation of 4, not 8

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -141,39 +145,47 @@ internal class HttpClient(
             headers.remove(Headers.CONTENT_LENGTH)
             contentLength = -1
         }
-
+        val inputStream = stethoHook?.interpretResponseStream(connection.inputStream)
+                ?: connection.inputStream
```
SleeplessByte: Please use indentation of 4, not 8

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -141,39 +145,47 @@ internal class HttpClient(
...
+        stethoHook?.postConnect()
```
SleeplessByte: I prefer `hooks.postConnect`.

Are you sure this is where this should be called? Headers are already retrieved. Maybe it's meant as post received `Connection` which happens to have the headers, then it is fine. I just don't know about it 😁

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -141,39 +145,47 @@ internal class HttpClient(
...
-                            shouldDecode = shouldDecode
-                        ),
-                        onProgress = { readBytes ->
-                            request.responseProgress(readBytes, contentLength ?: readBytes)
-                        }
-                    ).buffered(FuelManager.progressBufferSize)
-                },
-                { contentLength ?: -1 }
-            )
+                url = request.url,
```
SleeplessByte: Please reset the indentation for this entire block

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -141,39 +145,47 @@ internal class HttpClient(
...
 
-    private fun safeDataStream(connection: HttpURLConnection): InputStream? {
+    private fun safeDataStream(connection: HttpURLConnection, inputStream: InputStream): InputStream? {
         return try {
-            (connection.errorStream ?: connection.inputStream) ?.let { BufferedInputStream(it) }
+            BufferedInputStream((connection.errorStream ?: inputStream))
         } catch (exception: IOException) {
             // Stream error
-            try { (connection.errorStream ?: connection.inputStream).close() } catch (_: IOException) {}
+            stethoHook?.httpExchangeFailed(exception)
```
SleeplessByte: I prefer `hooks.onHttpError(exception)`

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -141,39 +145,47 @@ internal class HttpClient(
...
+            try {
+                (connection.errorStream ?: connection.inputStream).close()
+            } catch (_: IOException) {
+                stethoHook?.httpExchangeFailed(exception)
```
SleeplessByte: I don't think you need this. This happens when closing the stream fails, not when an HttpExchange fails.

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -239,17 +251,21 @@ internal class HttpClient(
...
         // The input and output streams returned by connection are not buffered. In order to give consistent progress
         // reporting, by means of flushing, the output stream here is buffered.
         body.writeTo(
-            ProgressOutputStream(
-                connection.outputStream,
-                onProgress = { writtenBytes ->
-                    request.requestProgress(writtenBytes, totalBytes ?: writtenBytes)
-                }
-            ).buffered(FuelManager.progressBufferSize)
+                ProgressOutputStream(
```
SleeplessByte: Please reset the indentation of this entire block

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -260,6 +276,13 @@ internal class HttpClient(
         Method.DELETE, Method.POST, Method.PUT, Method.PATCH -> connection.doOutput = true
     }
 
+    interface StethoHook {
```
SleeplessByte: I prefer `Hooks`

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -141,39 +145,47 @@ internal class HttpClient(
             headers.remove(Headers.CONTENT_LENGTH)
             contentLength = -1
         }
-
+        val inputStream = stethoHook?.interpretResponseStream(connection.inputStream)
```
SleeplessByte: I prefer something along the lines of `wrapResponseStream` as it's more clear what the intent is of the function

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -260,6 +276,13 @@ internal class HttpClient(
         Method.DELETE, Method.POST, Method.PUT, Method.PATCH -> connection.doOutput = true
     }
 
+    interface StethoHook {
+        fun preConnect(connection: HttpURLConnection, request: Request)
```
SleeplessByte: Missing kdoc

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -260,6 +276,13 @@ internal class HttpClient(
...
+        fun interpretResponseStream(inputStream: InputStream): InputStream
```
SleeplessByte: Missing kdoc

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -260,6 +276,13 @@ internal class HttpClient(
...
+        fun postConnect()
```
SleeplessByte: Missing kdoc

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -260,6 +276,13 @@ internal class HttpClient(
...
+        fun httpExchangeFailed(exception: IOException)
```
SleeplessByte: Missing kdoc

fuel-stetho/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,2 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.github.worker8.fuel.stetho" />
```
SleeplessByte: This seems the incorrect package

fuel-stetho/src/main/java/com/github/worker8/fuel/stetho/StethoHook.kt
```diff
@@ -0,0 +1,29 @@
+package com.github.worker8.fuel.stetho
```
SleeplessByte: This seems the incorrect package

sample/src/main/kotlin/com/example/fuel/MainActivity.kt
```diff
@@ -28,7 +28,9 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.launch
 import java.io.File
+
 import java.io.Reader
+import com.github.worker8.fuel.stetho.StethoHook
```
SleeplessByte: This seems the incorrect package

kittinunf:
@worker8 Thanks for your awesome work! Can you also rebase once #524 is merged so that you can test with some other methods aside from `GET`.

worker8:
Rebased, did some bug fix, and remove unnecessary spacing for easier review (sorry for that 😳 ).

----

### Some explanations to your questions
#### 1. Question
> It seems that the stethoHook usage is isolated to HttpClient and can be injected after it has been created. I don't think we need this field at all. -- SleeplessByte

#### 1. Answer
I'm implementing the 4 methods from `HttpClient.kt` in order to intercept the calls:

```
StethoManager#preConnect(connection: HttpURLConnection, requestEntity: SimpleRequestEntity?)
StethoManager#interpretResponseStream(inputStream: InputStream)
StethoManager#postConnect()
StethoManager#httpExchangeFailed(ioException: IOException)
```

So according to the docs of each methods, 
**preConnect(connection: HttpURLConnection, requestEntity: SimpleRequestEntity?)
S** ([reference](https://github.com/facebook/stetho/blob/6d148f20b96ecb29320cced6617830eda8af0426/stetho-urlconnection/src/main/java/com/facebook/stetho/urlconnection/StethoURLConnectionManager.java#L72 )):

> Indicates that the {@link HttpURLConnection} instance has been configured and is about
>   * to be used to initiate an actual HTTP connection.  Call this method before any of the
>   * active methods such as {@link HttpURLConnection#connect()},

It says "Call this method before any active methods", since we don't use `HttpURLConnection#connect()`, the `inputStream` is an active method. So it has to be placed before it.

For **postConnect()**([reference](https://github.com/facebook/stetho/blob/6d148f20b96ecb29320cced6617830eda8af0426/stetho-urlconnection/src/main/java/com/facebook/stetho/urlconnection/StethoURLConnectionManager.java#L89)):

> Indicates that the {@link HttpURLConnection} has just successfully exchanged HTTP messages
>   * (request headers + body and response headers) with the server but has not yet consumed
>   * the response body.

The doc says that it has to be placed in **before** the consumption of the response body and **after** the actual network call.

So I think we need to have hook inside the HttpClient, we cannot do it after 

---
#### 2. Question
> I prefer hooks.preConnect, where hooks takes multiple hooks, not just one. -- SleeplessByte

> I prefer Hooks -- SleeplessByte

#### 2. Answer 
The reason I make it `stethoHook` is because I don't know if the hook placement will be useful for anything else. I decided the location to place the hook solely based on it seems to be just very specific placement of the need of [StethoURLConnectionManager](https://github.com/facebook/stetho/blob/6d148f20b96ecb29320cced6617830eda8af0426/stetho-urlconnection/src/main/java/com/facebook/stetho/urlconnection/.java
).

I spend half a day moving the locations around of this hooks to make it work 😂 Just so you know, I don't actually fully understand why the hooks have to be placed at those locations... those are just the locations that I manage to get it working 😳

What I'm trying to say is that, if this is not useful for anything else, then I should name it as `stethoHook`, so that it is **clear** of what it is used for. 

However, I don't have enough knowledge about Fuel and HttpClient.kt, if you still think it's a useful `hook` or it aligns with your future plans of making hooks after reading this paragraph, then I'm happy to rename it! 👍 

--- 

3. Question
> I prefer something along the lines of wrapResponseStream as it's more clear what the intent is of the function -- SleeplessByte

> I prefer hooks.onHttpError(exception) -- SleeplessByte

3. Answer
Similar to 2., I name the method by copying the method name of [StethoURLConnectionManager#interpretResponseStream](https://github.com/facebook/stetho/blob/6d148f20b96ecb29320cced6617830eda8af0426/stetho-urlconnection/src/main/java/com/facebook/stetho/urlconnection/StethoURLConnectionManager.java#L130). I will rename this is you decide to name it `hook`.

Same for `onHttpError`, but we don't actually need this as per your other comment.

---

### There's a bug
Currently there's a bug though, when I try to run the demo activity to check my stetho implementation, I found this.

I commented all the methods except for `httpGet()`.

```
private fun execute() {
        httpGet()
//        httpPut()
//        httpPost()
//        httpDelete()
...same
```

And inside `httpGet()`, I need to add a delay like this:

```
private fun httpGet() {
        Fuel.get("/get", listOf("foo" to "foo", "bar" to "bar"))
                .also { Log.d(TAG, it.cUrlString()) }
                .responseString { _, _, result -> update(result) }
       Thread.sleep(1000) // <---- DELAY
        "/get"
                .httpGet()
                .also { Log.d(TAG, it.cUrlString()) }
                .responseString { _, _, result -> update(result) }
    }
```

**Without** the delay, I will **not** get any response in Stetho:

image

With the delay, everything runs fine:

image

It feels like a race condition, do you have any rough idea why this happens? 🤔

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
iNoles: I think you called postConnect too late in the process.

fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
```diff
@@ -27,8 +27,10 @@ import kotlin.coroutines.resumeWithException
 import kotlin.coroutines.suspendCoroutine
 import kotlin.math.max
 
-internal class HttpClient(
+class HttpClient(
     private val proxy: Proxy? = null,
+    var stethoHookFactory: (() -> StethoHook)? = null,
+    var stethoHook: StethoHook? = null,
```
iNoles: is this is still needed since you going to assign it from the Factory?

worker8: thanks! I removed it and make it a local variable.

SleeplessByte:
> So I think we need to have hook inside the HttpClient, we cannot do it after
Ah yes, I _agree_ with you that it has to be called _inside_, but I think having generic hooks is more powerfull, which leads to the next one:

> What I'm trying to say is that, if this is not useful for anything else, then I should name it as stethoHook, so that it is clear of what it is used for.
Completely agreed. I already have debugging purposes for the hooks right away, so I think we can generify it. So you basically hit the nail on its head! I **love** that you're doing this work, because I _think_ others can use it and extend on it and build on top of it! 

For example: cookies could be done using a hook!

As per the bug: Probably not a race condition but calling the method on the wrong place. I'll happily investigate with you if that's what it takes?

Keep up the great work @worker8 !

worker8:
I have other focus now, I will reopen this PR when I have time to do a deeper dive into understanding how Stetho and Fuel work.

Closing for now 🙇

PR #542, opened 2018-11-21
Add option to use .editorConfig in CONTRIBUTING.md

worker8:
## Description

Currently in the CONTRIBUTING.md, there is only one guide to run the `ktlint` formatter, that is through `ktlint --install-git-pre-commit-hook`, so in this PR, I'm adding another small paragraph for using `.editorconfig`. That is because I want to have more visibility on how my code is being formatted rather than until commit time (using the pre commit hook).

## Type of change

- [x] This is just a documentation update

## How Has This Been Tested?

I'm using Android Studio from JetBrain which has out of the box support for `.editorconfig`. So by running `$curl -O https://raw.githubusercontent.com/shyiko/ktlint/master/.editorconfig`, my IDE formatter can pick up the style directly.

## Checklist:


SleeplessByte:
I think there is still an issue with undeterministic tests, so rebuilding. If it does, this is a ✔️ from me!

PR #543, opened 2018-11-23
[Common] Enable snapshot version

kittinunf:
## Description

As our discussion, it looks like it is gonna be beneficiary to everyone if we do support this *officially*. I think we will publish the `-SNAPSHOT` everytime we hit `master` branch.

## Type of change

Check all that apply

- [x] Chores

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/deploy_bintray.sh
+++ b/deploy_bintray.sh
@@ -14,3 +14,18 @@ if [[ "$TRAVIS_BRANCH" == */release-v* ]]; then
   done
 
 fi
+
+if [[ "$TRAVIS_BRANCH" == master ]]; then
+
+  echo "We're on master branch, deploying SNAPSHOT"
+
+  for i in $(ls -d */);
+  do
+    m=${i%%/}
+    if [[ $m == fuel* ]]; then
+      echo ">> Deploying $m ..."
+      ./gradlew :$m:clean :$m:build :$m:bintrayUpload -PBINTRAY_USER=$BINTRAY_USER -PBINTRAY_KEY=$BINTRAY_KEY -PdryRun=false -Ppublish=true
+    fi
+  done
+
+fi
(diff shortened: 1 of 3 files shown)


deploy_bintray.sh
```diff
@@ -14,3 +14,18 @@ if [[ "$TRAVIS_BRANCH" == */release-v* ]]; then
...
+if [[ "$TRAVIS_BRANCH" == master ]]; then
+
+  echo "We're on master branch, deploying SNAPSHOT"
+
+  for i in $(ls -d */);
+  do
+    m=${i%%/}
+    if [[ $m == fuel* ]]; then
+      echo ">> Deploying $m ..."
+      ./gradlew :$m:clean :$m:build :$m:bintrayUpload -PBINTRAY_USER=$BINTRAY_USER -PBINTRAY_KEY=$BINTRAY_KEY -PdryRun=false -Ppublish=true
```
iNoles: I think we should exclude test and sample here

kittinunf: I think it should be okay because test and sample module don't have `bintrayUpload` configured.

kittinunf:
Shall we merge this?

SleeplessByte:
Don't we have that issue with 10GB limit etc? Or what were the concerns exactly?

kittinunf:
I am not sure though. But I can write a script to prune them if we hit the space limit.

kittinunf:
It looks like we couldn't find a good way for this, maybe I will close it for now.

NikkyAI:
https://jitpack.io/#kittinunf/Fuel is still working just fine
just use `group = "com.github.kittinunf.Fuel", name = subproject, version = "-SNAPSHOT"`
subprojects owuld be `fuel`, `fuel-coroutines`, etc

PR #550, opened 2018-11-28
Restrict body printing and normalize for Request/Response

SleeplessByte:
## Description

This whitelists text media types as well as known application text representations to display when logging the body. Anything else, including custom and vendor media types that are not explicitly text are displayed as (x bytes of y).

- `text/*` media types
- media types that end with `+json` or `+xml` (registered text extensions)
- the following `application/` media types:
  - `application/csv`
  - `application/javascript`, `application/typescript`, `application/vnd.coffeescript`
  - `application/json`
  - `application/xml`
  - `application/x-yaml`

It's not really extensible, but hopefully someone can make it so in a future PR.

It explicitly tests the types listed here, as well as correct "summary" representation of
- `image/bmp`, `image/gif`, `image/jpeg`, `image/png`, `image/tiff`, `image/webp`, `image/x-icon`
- `audio/aac`, `audio/midi`, `audio/x-midi`, `audio/ogg`, `audio/wav`, `audio/webm`, `audio/3gpp`, `audio/3gpp2`
- `video/mpeg`, `video/ogg`, `video/webm`, `video/x-msvideo`, `video/3gpp`, `video/3gpp2`
- `font/otf`, `font/ttf`, `font/woff`, `font/woff2`
- `application/octet-stream`

The reason it's testing for all these is that someone might change the implementation, but we still want this behaviour to be preserved.

Should fix issues with the types above as well as binary custom / vendor / unregistered media types.

## Type of change

Check all that apply

- [X] Bug fix (a non-breaking change which fixes an issue)
- [X] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] Refactoring (a change which changes the current internal or external interface)
- [ ] This change requires a documentation update

## How Has This Been Tested?

New tests have been added.

## Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation, if necessary
- [X] My changes generate no new compiler warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [ ] Inspect the bytecode viewer, including reasoning why

PR #553, opened 2018-11-28
Parameter encoding for all requests

### LINKED ISSUE #546: No easy way to add parameters to query for POST
# Feature Request

## Description
Need to be able to add parameters to the URL of a POST request that has a JSON body. The only way to do it is for me to encode the query into the path string.

## Proposed Solution
A way to have parameters for a POST to put the parameters into the URL and not assume that parameters automatically means form data where the body contains the query.

## Alternatives I've considered
Going to the extra work to build query string into the path string..

## Additional context
I don't control the API I am talking to and barely even have any documentation so I cannot investigate alternatives to using parameters in the URL for this API.

SleeplessByte:
## Description

This enables encoding parameters in urls correctly for ALL request types, including array parameters, null parameters, empty parameters.

Also allows for #546 without any extra configuration. Try to put the parameters in the body, but if there's already a body, put it in the URL. Still allows for "pre-encoded" urls.

Fixes #546

## Type of change

Check all that apply

- [X] New feature (a non-breaking change which adds functionality)

## How Has This Been Tested?

New tests have been added to test this behaviour

## Checklist:


SleeplessByte:
Based on #550 . Merge that first.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/ParameterEncoder.kt
```diff
@@ -0,0 +1,79 @@
...
+            ((values as? Iterable<*>)?.toList() ?: (values as? Array<*>)?.toList())?.let {
+                val encodedKey = "${URLEncoder.encode(key, "UTF-8")}[]"
+                it.map { value -> encodedKey to URLEncoder.encode(value.toString(), "UTF-8") }
+
+            // Deal with regular
+            } ?: values?.let {
+                listOf(URLEncoder.encode(key, "UTF-8") to URLEncoder.encode(values.toString(), "UTF-8"))
+
+            // Parameters that are null are stripped
+            } ?: emptyList()
```
kittinunf: [nits]  Shall we eagerly filter the null value out first? Then, I think we can just deal with less `?.` thingy

``` Kotlin
parameters.filterNot { (key, values) -> values == null }
.flatMap {
   // as current impl
}
```

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/ParameterEncoder.kt
```diff
@@ -0,0 +1,79 @@
...
+            if (contentType?.startsWith("multipart/form-data") == true) {
+                return@inner next(request)
+            }
+
+            // If it can be added to the body
+            if (request.body.isEmpty() && allowParametersInBody(request.method)) {
+                if (contentType.isNullOrBlank() || contentType.startsWith("application/x-www-form-urlencoded")) {
+                    return@inner next(
+                        request
+                            .header(Headers.CONTENT_TYPE, "application/x-www-form-urlencoded")
```
kittinunf: [Q] I am confused on this line a bit why do we need to add "application/x-www-form-urlencoded" after the check once again?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/ParameterEncoder.kt
SleeplessByte: Yes!

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/ParameterEncoder.kt
SleeplessByte: We want to support the following:

```kotlin
requestWithParams
  .header(Headers.CONTENT_TYPE, "application/x-www-form-urlencoded")
  .await()
```

... but also ...

```
requestWithParams
  .await()
```

Both cases are the same, it needs to encode the parameters into the `body`; but in the second case the header might be `null` or `blank` and still need to be set. We don't want users who opt to set it explicitly to be fucked and have their parameters not encoded 📦

kittinunf: AHHH I see what it is now.

SleeplessByte:
Do you like?

kittinunf:
@SleeplessByte Yes! I think after the small nits being fixed, I am happy to see this merged.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/interceptors/ParameterEncoder.kt
SleeplessByte: Done!

PR #554, opened 2018-11-29
Don't need to Run Samples in Travis CI

iNoles:
## Description
Sample is like show developers what is like to run Fuel.

## Type of change

Check all that apply

## How Has This Been Tested?

## Checklist:


settings.gradle.kts
```diff
@@ -32,4 +32,3 @@ include(":fuel-kotlinx-serialization")
 include(":fuel-moshi")
 include(":fuel-test")
 include(":fuel-json")
-include(":sample")
```
kittinunf: [Q] By doing this it means that it won't recognize as a module in IntelliJ? Is this the intended effect?

SleeplessByte: I don't think so

iNoles: Intellij IDEA able to search everything under new source set. It should detect sample.

SleeplessByte:
It adds it back to settings it seems

iNoles:
Closed this for better PR.

PR #559, opened 2018-12-04
 Overhaul upload requests, Blob and DataPart

### LINKED ISSUE #344: How to upload customed Blobs (like DataPart) in multipart way?

### LINKED ISSUE #390: Upload list of files as single element in multipart body
See the issue https://github.com/kittinunf/Fuel/issues/169, which seams try to get the same, but was closed before somebody give the solution.

I tried solution in that issue, but the resulting multipart body still contains separate part for each file. The name of the parts are "files1", "files2", etc.
What I need (and seams the originator of that topic) the multipart with single part named "files" and containing the list of files.

How can we do THAT?

### LINKED ISSUE #399: multipart/form-data boundary isn't set correctly

### LINKED ISSUE #479: Improve multipart features for httpPost and httpUpload

### LINKED ISSUE #513: [fuel-core] Overwrite header in case of Content-Type: multipart/form-data
# Feature Request

## Description
This feature solves a problem I've had when constructing a Multipart/form-data request.

If one sets the Content-Type header to multipart/form-data manually (done in the snippet below) this request wont work properly, since the boundry value in the header, pointing out the file, is missing.

If the manually set content-type is removed, the request works fine.

```
Fuel.upload(path = baseUrl, method = Method.POST)
                .header(mutableMapOf(
                    "apikey" to key,
                    "Accept" to "application/json",
                    "Cache-Control" to "no-cache"
                    "Content-Type" to "multipart/form-data"
                ))
                .dataParts { _, _ -> listOf(DataPart(file, "file", "image/jpeg")) }
                .responseJson { _, _, result ->
                    result.fold(
                            success = { data -> cb(data.content) },
                            failure = { error ->
                                println("An error of type ${error.exception} happened: ${error.message}")
                                cb("")
                            }
                    )
                }
```

## Proposed Solution
Fuel should in my opinion in the case of such a specific request, *overwrite* the header even though a user beforehand set it manually, since the request otherwise wont work at all.

SleeplessByte:
## Description

Based on RFC7578: https://tools.ietf.org/html/rfc7578

This is a breaking change, but mostly deprecates current functions as error.

Removes all the extranous methods and bookkeeping from UploadRequest and UploadBody, and delegates those to the DataPart. This also allows for InlineDataPart that yield a file on the server and removes automatic naming for dataparts (_1, _2), because the correct way to put multiple files under the same name is to give them... the same name!

Fixes #344
Fixes #390
Fixes #399
Fixes #479
Fixes #513

## Type of change

Check all that apply

- [X] New feature (a non-breaking change which adds functionality)

## How Has This Been Tested?

Tests have been updated and new ones have been added

> Note that due to deprecation this _DOES_ generate compiler warnings, but these are expected.

## Checklist:


fuel/README.md
```diff
@@ -390,60 +390,69 @@ Fuel supports multipart uploads using the `.upload()` feature. You can turn _any
...
 
 ```kotlin
-val formData = listOf("Email" to "mail@example.com", "Name" to "Joe Smith" )
-
-Fuel.upload("/post", param = formData)
-    .dataParts { request, url -> listOf<DataPart>() }
+Fuel.upload("/post")
+    .add(
+        FileDataPart(File("myfile.json"), name = "file", fileName="contents.json"),
+        InlineDataPart(myinlinecontent, name = "metadata", fileName="metadata.json", contentType = "application/json")
```
markGilchrist: [nits] should `myinlinecontent` be `myInlineContent` ?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/DataPart.kt
```diff
@@ -1,9 +1,223 @@
...
+ *    // Content-Disposition: "form-data; name=field-name[]; filename=baz.json"
+ *    // Content-Type: "application/json"
+ *    //
+ *    // <file-contents baz.json>
+ *
+ *  @param file [File] the source file
+ *  @param name [String] the field name, defaults to the file name without extension
+ *  @param fileName [String] the remote file name, or null
+ *  @param contentType [String] the content type of the file contents; defaults to a guess using [guessContentType]
+ *  @param contentDisposition [String] defaults to form-data with name and filename set, unless filename is null
```
markGilchrist: @SleeplessByte  `the Shakespeare of kDocs` !

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/UploadBody.kt
```diff
@@ -1,19 +1,28 @@
...
-import java.net.URLConnection
 import java.nio.charset.Charset
 
+internal class BoundaryMissing(request: UploadRequest) : FuelError(
+    IllegalArgumentException(
+        "The Request is missing the boundary parameter in its Content-Type.\n\n" +
+        "This can happen if you manually overwrite the Content-Type but forget to set a boundary. The boundary is \n" +
+        "normally set automatically when you call \"request.upload()\". Remove manually setting the Content-Type or \n" +
+        "add the boundary parameter to the Content-Type for this request: \n\n" +
+            "\trequest.header(Headers.ContentType, \"multipart/form-data; boundary=custom-boundary\")"
```
markGilchrist: great

fuel/src/test/kotlin/com/github/kittinunf/fuel/BlockingRequestTest.kt
```diff
@@ -290,12 +291,8 @@ class BlockingRequestTest : MockHttpTestCase() {
 
         val (request, response, data) =
                 manager.upload(mock.path("upload"), parameters = listOf("foo" to "bar", "foo1" to "bar1"))
-                        .source { _, _ ->
-                            val dir = System.getProperty("user.dir")
-                            val currentDir = File(dir, "src/test/assets")
-                            File(currentDir, "lorem_ipsum_long.tmp")
-                        }
-                        .responseString()
+                    .add { FileDataPart(File(File(System.getProperty("user.dir"), "src/test/assets"), "lorem_ipsum_long.tmp")) }
```
kittinunf: We have `File(File` here sir!

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/DataPart.kt
```diff
@@ -1,9 +1,223 @@
...
+ */
+data class FileDataPart(
     val file: File,
```
kittinunf: [Question] As this is a `FileDataPart`, do you think we can add another API in such a way that we can skip passing `File` object altogether for the sake of ease of use? For example, In the test the usage could be

```kotlin
request.add(FileDataPart(System.getProperty("user.dir").resolve("src/test/assets"), "lorem_ipsum_long.tmp"))
```

I think the usage of passing a "String" or a path that indicates the location of the file should be the most ubiquitous?

fuel/src/test/kotlin/com/github/kittinunf/fuel/BlockingRequestTest.kt
SleeplessByte: Hah. That's a woooopsiedaisy.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/DataPart.kt
kittinunf: Ah, I think that is a reasonable approach! 👍

fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestPathStringConvertibleExtensionTest.kt
```diff
@@ -151,11 +152,7 @@ class RequestPathStringConvertibleExtensionTest : MockHttpTestCase() {
 
         val (request, response, result) = PathStringConvertibleImpl(mock.path("http-upload"))
             .httpUpload()
-            .source { _, _ ->
-                val dir = System.getProperty("user.dir")
-                val currentDir = File(dir, "src/test/assets")
-                File(currentDir, "lorem_ipsum_long.tmp")
-            }
+            .add { FileDataPart(File(File(System.getProperty("user.dir"), "src/test/assets"), "lorem_ipsum_long.tmp")) }
```
kittinunf: Here too sir!

fuel/src/test/kotlin/com/github/kittinunf/fuel/BlockingRequestTest.kt
SleeplessByte: I changed it to

```kotlin
FileDataPart.from(
  File(System.getProperty("user.dir")).resolve("/src/test/assets"), 
  "lorem_ipsum_long.tmp"
)
```

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/DataPart.kt
SleeplessByte: This now works.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/DataPart.kt
```diff
@@ -1,9 +1,259 @@
...
+         * @return [FileDataPart] the DataPart
+         */
+        fun from(directory: File, filename: String, name: String? = null, remoteFileName: String? = filename, contentType: String? = null): DataPart {
+            val file = File(directory, filename)
+            return FileDataPart(
+                file = file,
+                name = name ?: file.nameWithoutExtension,
+                filename = remoteFileName,
+                contentType = contentType ?: guessContentType(file)
+            )
```
kittinunf: **[ktlint]** <sub>reported by [reviewdog](https://github.com/haya14busa/reviewdog) :dog:</sub>
Unexpected blank line(s) before "}"

fuel/src/test/kotlin/com/github/kittinunf/fuel/BlockingRequestTest.kt
```diff
@@ -288,14 +289,11 @@ class BlockingRequestTest : MockHttpTestCase() {
...
+        val path = File(System.getProperty("user.dir"), "/src/test/assets").resolve("lorem_ipsum_long.tmp").path
         val (request, response, data) =
                 manager.upload(mock.path("upload"), parameters = listOf("foo" to "bar", "foo1" to "bar1"))
-                        .source { _, _ ->
-                            val dir = System.getProperty("user.dir")
-                            val currentDir = File(dir, "src/test/assets")
-                            File(currentDir, "lorem_ipsum_long.tmp")
-                        }
-                        .responseString()
+                    .add { FileDataPart.from(path,  "lorem_ipsum_long.tmp") }
```
kittinunf: **[ktlint]** <sub>reported by [reviewdog](https://github.com/haya14busa/reviewdog) :dog:</sub>
Unnecessary space(s)

PR #563, opened 2018-12-16
[fuel-core] Add Response validator to ExecutionOption

### LINKED ISSUE #379: API Request: graceful handling of 304
Linking the old ticket #191,

When using `Fuel` and setting `If-None-Match` manually, when the server returns a 304, currently an `HttpException` is raised, whereas this is actually expected. The happy-path shouldn't be "exceptional". When a 204 (No Content) is returned, Fuel doesn't complain unless you try to deserialize. 

As it stands, by default `redirectResponseInterceptor` will not match the 304 (which is correct), followed by the `validatorResponseInterceptor` will error as it's not included in the default `validRange`. Only the 200-299 range is handled as valid, whereas 100-399 should be considered valid; only 400-599 are error codes.
https://github.com/kittinunf/Fuel/blob/56cf7ce2af6ea39b2fb013564d188a812c42cf55/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/FuelManager.kt#L58

The default should not indicate something is an error when it's not (304, 300, 103, ...).
RxJS used to have the same issue, but this was fixed: https://github.com/ReactiveX/rxjs/issues/3308

kittinunf:
## Description

This PR aims to fix #379. Invalidate usage of `validatorResponseInterceptor` and introducing the `responseValidator` in the `ExecutionOptions` instead. 

### Usage 
It should be pretty straightforward. The default implementation is to check whether it is fallen into 4xx-5xx or not, if yes then it will be considered failure. And it could be overridden just like other things in `RequestExecutionOptions` .

## Type of change

Check all that apply

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

## Checklist:


kittinunf:
@SleeplessByte Can you please help check whether it is a good direction to go? Maybe I misunderstand something ...

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/DefaultRequest.kt
```diff
@@ -198,8 +199,8 @@ data class DefaultRequest(
      */
     override fun body(openStream: BodySource, calculateLength: BodyLength?, charset: Charset, repeatable: Boolean): Request {
         _body = DefaultBody
-            .from(openStream = openStream, calculateLength = calculateLength, charset = charset)
-            .let { body -> if (repeatable) body.asRepeatable() else body }
+                .from(openStream = openStream, calculateLength = calculateLength, charset = charset)
```
SleeplessByte: You probably didn't want to change this indentation to 8 instead of 4.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/DefaultRequest.kt
```diff
@@ -217,7 +218,7 @@ data class DefaultRequest(
      * @return [Request] the request
      */
     override fun body(stream: InputStream, calculateLength: BodyLength?, charset: Charset, repeatable: Boolean) =
-        body(openStream = { stream }, calculateLength = calculateLength, charset = charset, repeatable = repeatable)
+            body(openStream = { stream }, calculateLength = calculateLength, charset = charset, repeatable = repeatable)
```
SleeplessByte: You probably didn't want to change this indentation to 8 instead of 4.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/DefaultRequest.kt
```diff
@@ -227,7 +228,7 @@ data class DefaultRequest(
      * @return [Request] the request
      */
     override fun body(bytes: ByteArray, charset: Charset) =
-        body(stream = ByteArrayInputStream(bytes), calculateLength = { bytes.size.toLong() }, charset = charset, repeatable = true)
+            body(stream = ByteArrayInputStream(bytes), calculateLength = { bytes.size.toLong() }, charset = charset, repeatable = true)
```
SleeplessByte: You probably didn't want to change this indentation to 8 instead of 4.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/DefaultRequest.kt
```diff
@@ -237,13 +238,13 @@ data class DefaultRequest(
...
      */
     override fun body(body: String, charset: Charset): Request =
-        body(bytes = body.toByteArray(charset), charset = charset)
-            .let {
-                if (header(Headers.CONTENT_TYPE).lastOrNull().isNullOrBlank())
-                    header(Headers.CONTENT_TYPE, "text/plain; charset=${charset.name()}")
-                else
-                    it
-            }
+            body(bytes = body.toByteArray(charset), charset = charset)
```
SleeplessByte: You probably didn't want to change this indentation to 8 instead of 4.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/DefaultRequest.kt
```diff
@@ -392,45 +406,62 @@ data class DefaultRequest(
     }
 
     override fun response(handler: ResponseResultHandler<ByteArray>) =
-        response(ByteArrayDeserializer(), handler)
+            response(ByteArrayDeserializer(), handler)
```
SleeplessByte: You probably didn't want to change this indentation to 8 instead of 4.

Here and below.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/DefaultRequest.kt
kittinunf: I think we need to sync on this! Can you share your setting and push to the project?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/DefaultRequest.kt
kittinunf: Roger that

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/FuelManager.kt
```diff
@@ -65,7 +64,7 @@ class FuelManager : RequestFactory, RequestFactory.Convenience {
     private val requestInterceptors: MutableList<FoldableRequestInterceptor> =
             mutableListOf(ParameterEncoder)
     private val responseInterceptors: MutableList<FoldableResponseInterceptor> =
-            mutableListOf(redirectResponseInterceptor(this), validatorResponseInterceptor(200..299))
```
kittinunf: I don't feel like using `validatorResponseInterceptor` is a good use of Interceptor anymore. I think I will just take this opportunity and remove it without effect users' behaviors.

iNoles: range from 200 to 299 can be difficult if there is more http status code that considered to be okay.

kittinunf: Yeah, agree. After this PR is merged, `!(response.isServerError || response.isClientError)` is the new logic, so it should be okay.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/requests/DefaultRequest.kt
SleeplessByte: Yes, but I think I just followed the klint instructions. I think the contributors file also has a few lines on how to get editorconfig working.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/RequestExecutionOptions.kt
```diff
@@ -29,6 +31,9 @@ data class RequestExecutionOptions(
     var allowRedirects: Boolean? = null
     var useHttpCache: Boolean? = null
     var interruptCallbacks: MutableCollection<InterruptCallback> = mutableListOf()
+    var responseValidator: ResponseValidator = { response ->
+        !(response.isServerError || response.isClientError)
```
kittinunf: This should allow 3xx as a Success in the `Result` value.

kittinunf:
@SleeplessByte Happy new year bro 🎉 

One more check when possible! 🙏

kittinunf:
Shall we merge this? If anyone can review? 🙏

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/RequestExecutionOptions.kt
SleeplessByte: Haha, did you review yourself 💯

SleeplessByte:
The only question I have left @kittinunf , is what happens in this case:

```
val request = Fuel.get(url)
request['If-None-Match'] = 'some etag'

val response = request.responseObject(...)
// This will be fine if the etag doesn't match (the resource has changed) and you get a 200 or the like

// What happens if it DOES match and you get a 304? 
// - Do we throw? 
// - Could we return an empty body -- is there a nice way we can handle that?
// - Do we need a new API that basically tells you if you got "responseWithBodyOfType, responseWithoutBody"?
```

kittinunf:
Let's write a test for that ⬆️ 🙃

SleeplessByte:
🤗

kittinunf:
@SleeplessByte Can we defer the abovementioned for later fix? So that we can merge this and release 2.0

SleeplessByte:
@kittinunf sure, just open up an issue, and then merge. Put it in the release notes :)

kittinunf:
@SleeplessByte yeah, I created here #578. Can you approve this if there is no more concern aside from that?

PR #602, opened 2019-02-24
[Bug] [Stetho] Fix bug in stetho module

kittinunf:
## Description

This PR attempts to fix but in Stetho module, where it only works on the first request but not on the subsequent request.

## Type of change

The reason behind this is the `StethoURLConnectionManager` is designed to be used only once and throw away according to the documentation. Since our `Stetho` hook is being created once, the current implementation retains the same `StethoURLConnectionManager` instance which means that it makes the subsequent requests unable to inspect because `StethoURLConnectionManager` became unusable at the point.

### Implementation

After I carefully think about a good way to do this, it is clear that we need some sort of "in-memory" to cache the pair of `Request` => `StethoURLConnectionManager` so that it could be created/used/disposed according to the correct request. Since the current Request class is lagging some sort of "tagging" system, it could prove to be challenging to find the correct `StethoURLConnectionManager` according to the Request.

This PR has 2 main goals.

1. Adding the tagging mechanism into `Request` class
2. Improve the `Stetho` hook implementation

### New features

In this PR, I have added tagging feature for Request object;

```kotlin
fun tag(t: Any): Request

fun <T : Any> getTag(clazz: KClass<T>): T?
```

The top interface is to allow a user to be able to add arbitrarily object into the internal `hashMap`. The second one is to allow a user to be able to retrieve a previously tagged object by using `KClass<T>` as key.

### Bug fix

With abovementioned implementation, `StethoHook` uses the `randomUUID` to attach to the request. Given that the same request should call in all of the hook functions, the correct UUID can be retrieved and call at the appropriate timing. 

Check all that apply

## How Has This Been Tested?

## Checklist:

- [x] Inspect the bytecode viewer, including reasoning why

PS. Thanks from report from @worker8


kittinunf:
Yes, I think the `Request` object cannot be used as it will be wrapped and changed. I think it will be future-proof if we have something that is more static and easily used as a map key.

> And to that question: what happens if it redirects? Shouldn't UUID be forwarded? Or maybe linked?

💡 🚨 You are extremely right!!!! I don't think this would work ....

SleeplessByte:
I think that if you copy over the tag in the RedirectionInteractor, this can be merged @kittinunf

kittinunf:
@SleeplessByte 

It looks like this code

```kotlin
Fuel.get("/redirect-to", listOf("url" to "/get"))
                .responseString { _, _, result -> update(result) }
```

Produces this

screen shot 2019-03-02 at 16 28 24

I guess it looks fine as it is @SleeplessByte WDYT?

SleeplessByte:
@kittinunf yeah, in reality they ARE two requests, so as long as everything else works it's fine with me. Feel free to merge :)

kittinunf:
Thanks both of you for your reviews :)

PR #603, opened 2019-02-26
[Bug] Fix coroutine bug

### LINKED ISSUE #601: -1 returned for status code when using await with coroutines
# Bug Report

## Description
When my request is unsuccessful, and is awaited using the coroutine methods, the status code is being returned as -1

```Kotlin
request.responseString().second.statusCode // RETURNS 401
request.awaitStringResponseResult().second.statusCode // RETURNS -1
```

## Expected behavior
The correct status code should be returned when using coroutines. -1 is not a valid status code.

### Development Machine
_Complete the following information if applicable_

 - OS: Windows 10
 - IDE: IntelliJ
 - Fuel version: 2.0.1
 - Kotlin version: 1.3.21

```
implementation group: 'org.jetbrains.kotlinx', name: 'kotlinx-coroutines-core', version: '1.1.1'
implementation 'com.github.kittinunf.fuel:fuel:2.0.1'
implementation 'com.github.kittinunf.fuel:fuel-coroutines:2.0.1'
```

kittinunf:
## Description

Fixes #601 

The problem is this line `initialResult.getOrElse(Response.error())`. Since in the error case, the `initialResult` will be null then, it always return the empty version of `Response`. 

The correct implementation is

```kotlin
it.fold({ (response, _) -> response }, { error -> error.response })
```
where we get the `error.response` in the error case as well.

## Type of change

Check all that apply

## How Has This Been Tested?

We have test that is always emitting an error in the previous version.

``` kotlin
assertThat(data, notNullValue())
assertThat(error, notNullValue())
assertThat(error.statusCode, equalTo(404))
assertThat(error.isSuccessful, equalTo(false))
assertThat(error.headers["foo"], equalTo(listOf("bar") as Collection<String>))
```

Request from @SleeplessByte 

- [x] tests that second is a response with the correct error code
- [x] tests that third is a result with an error with a response with the correct error code
- [x] test that BOTH link to the same "bodyStream" (as in you can use both as error)

## Checklist:


fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
```diff
@@ -1,5 +1,8 @@
 package com.github.kittinunf.fuel.core
 
+import java.util.HashMap
```
kittinunf: I wonder how we can build without this .... 🤔

iNoles: Why not https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-hash-map/index.html?

iNoles: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-mutable-map/index.html

fuel-coroutines/src/test/kotlin/com/github/kittinunf/fuel/coroutines/ObjectTest.kt
```diff
@@ -67,9 +69,9 @@ class ObjectTest : MockHttpTestCase() {
     fun awaitObjectResponse() = runBlocking {
         try {
             val (request, response, data) = randomUuid().awaitObjectResponse(UUIDResponseDeserializer)
-            assertThat("Expected request to be not null", request, notNullValue())
-            assertThat("Expected response to be not null", response, notNullValue())
-            assertThat("Expected data to be not null", data, notNullValue())
```
SleeplessByte: The reason these messages are here is that in some context such as CI you only get `expected not null , actual null`. These messages change the output to be `expected xxx to be not null`.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
SleeplessByte: Why do we need this? We didn't need this before??

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
```diff
@@ -1,5 +1,8 @@
...
+import java.util.HashSet
```
SleeplessByte: Why do we need this? We didn't need this before??

kittinunf: Without this, the IDE shows red though .... and auto adds it for me.

fuel-coroutines/src/test/kotlin/com/github/kittinunf/fuel/coroutines/ObjectTest.kt
kittinunf: I think we can pin down to the line that throws Error right? I find it redundant ...

Anyway if you want this to stay, I will revert it back.

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
kittinunf: Fixed sir! a069508

kittinunf:
@SleeplessByte I have addressed your concern here 0de4ef6 !! 🙏

fuel-coroutines/src/test/kotlin/com/github/kittinunf/fuel/coroutines/ObjectTest.kt
SleeplessByte: > I think we can pin down to the line that throws Error right? I find it a bit redundant 🤔
Not on CI! Or am I wrong?

fuel/src/main/kotlin/com/github/kittinunf/fuel/core/Headers.kt
SleeplessByte: Weird?? Doesn't show that for me 🗡

PR #610, opened 2019-03-14
Fix body representation in case of `charset` in ContentType

kenrube:
## Description

Very often `Content-Type` header in responses contains `charset` and it fails body representation (e.g. when call `response.toString()`): instead of body's text representation we see `(XX bytes of application/json; charset=UTF-8)` for example. This fix adds to `TEXT_CONTENT_TYPE` regex support for [rich set of character sets](http://www.iana.org/assignments/character-sets/character-sets.xhtml).

## Type of change

## How Has This Been Tested?

Added 2 tests:
- for `application/json; charset=utf-8`
- for `application/csv; charset=utf-16be`

## Checklist:

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/BodyRepresentation.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/BodyRepresentation.kt
@@ -1,5 +1,7 @@
 package com.github.kittinunf.fuel.core
 
+import java.nio.charset.Charset
+
 private val TEXT_CONTENT_TYPE = Regex(
     "^(?:text/.*|application/(?:csv|javascript|json|typescript|xml|x-yaml|x-www-form-urlencoded|vnd\\.coffeescript)|.*\\+(?:xml|json))(; charset=.+)*$"
 )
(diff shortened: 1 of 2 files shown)


SleeplessByte:
Hi there @kenrube , 

This is a great addition. I don't actually understand why I've not run into this, because all our requests use charset=utf-8.

That said:
- If you make this change, you must capture the charset and apply it [here](https://github.com/kittinunf/fuel/blob/66995f6decae9c5ca303a50724fb39f75a8918c1/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/BodyRepresentation.kt#L17), otherwise the string will still look mangled.
- Please add at least one test that fails without your change and passes with your change. I think it'd be best to add a test for utf-8 and something else than utf-8

kenrube:
@SleeplessByte Ok, I'll do that a bit later

kittinunf:
Can we somehow test for this behavior though? 🤔

kenrube:
@SleeplessByte @kittinunf Fixed

kittinunf:
Wow, this looks sweet to me 🍬

SleeplessByte:
Thanks @kenrube . That is exactly what I was looking for.

PR #622, opened 2019-04-21
[Chore] Modernize project setup

kittinunf:
## Description

This PR aims to modernize the project gradle usage.

- [x] Remove all of `compile` and use `implementation` counterparts
- [x] Migrate Gradle to `5.2` instead of `4.10`
- [x] Fix inconsistencies around the projects (move files, take out JSON test from Android module etc.)

## Type of change

Check all that apply

## How Has This Been Tested?

## Checklist:


kittinunf:
Please ignore CircleCI build for now.

gradlew
```diff
@@ -28,7 +28,7 @@ APP_NAME="Gradle"
 APP_BASE_NAME=`basename "$0"`
 
 # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS=""
+DEFAULT_JVM_OPTS='"-Xmx64m"'
```
NikkyAI: i think this can be set in gradle.properties
and not be wiped on wrapper upgrade

kittinunf: it is updated automatically by using the wrapper update ...

kittinunf:
@NikkyAI Agree I will use `api` for `fuel-coroutine`

SleeplessByte:
Apart from Nikkys comments, lgtm

gradlew
kittinunf: revert!

kittinunf:
It looks like when updated to Gradle 5.2, it will break the test 🤔 🤦‍♂️

iNoles:
"To honour the JVM settings for this build a new JVM will be forked. Please consider using the daemon: https://docs.gradle.org/5.2/userguide/gradle_daemon.html."

kittinunf:
@iNoles I thought we set it here `org.gradle.daemon=true`, daemon should be enabled by default no?

sample/build.gradle.kts
```diff
@@ -8,20 +7,24 @@ plugins {
 }
 
 dependencies {
-    implementation(Kotlin.stdlib)
+    // fuel related libraries
```
iNoles: these should be implementation because api should be only in the libraries.

kittinunf: it is a sample app, so ... 🤷‍♂️ ww

kittinunf: Addressed!

PR #628, opened 2019-04-26
[CI] Another attempt of Circle CI

kittinunf:


kittinunf:
It looks like this doesn't work. Giving up.

PR #645, opened 2019-06-01
:sparkles: Add support for custom Gson instances

lostintime:
Add API support for passing custom Gson instances for serializing
request entities or deserializing response body

## Description

Current `fuel-gson` implementation have Gson instances hardcoded, I've added few overrides to existing functions to be able to pass a custom Gson instance.

For some of new functions - default values may be used (vs override), but I'm not sure which one is safer for binary compatibility, hopefully you may give me a hint here, ex:

```kotlin
inline fun <reified T : Any> Request.jsonBody(src: T, gson: Gson) = ...

inline fun <reified T : Any> Request.jsonBody(src: T) = this.jsonBody(src, Gson())
```

vs

```kotlin
inline fun <reified T : Any> Request.jsonBody(src: T, gson: Gson = Gson()) = 
```

## Type of change

Check all that apply

## Checklist:

### DIFF
--- a/fuel-gson/src/main/kotlin/com/github/kittinunf/fuel/gson/FuelGson.kt
+++ b/fuel-gson/src/main/kotlin/com/github/kittinunf/fuel/gson/FuelGson.kt
@@ -73,15 +73,11 @@ inline fun <reified T : Any> Request.responseObject(gson: Gson) =
  *
  * @return [ResponseDeserializable<T>] the deserializer
  */
-inline fun <reified T : Any> gsonDeserializerOf(clazz: Class<T>) = object : ResponseDeserializable<T> {
-    override fun deserialize(reader: Reader): T? = Gson().fromJson<T>(reader, clazz)
-}
+inline fun <reified T : Any> gsonDeserializerOf(clazz: Class<T>) = gsonDeserializer<T>()
 
-inline fun <reified T : Any> gsonDeserializer(gson: Gson) = object : ResponseDeserializable<T> {
+inline fun <reified T : Any> gsonDeserializer(gson: Gson = Gson()) = object : ResponseDeserializable<T> {
     override fun deserialize(reader: Reader): T? = gson.fromJson<T>(reader, object : TypeToken<T>() {}.type)
 }
-
-inline fun <reified T : Any> gsonDeserializer() = gsonDeserializer<T>(Gson())
 
 /**
  * Serializes [src] to json and sets the body as application/json
--- a/fuel-gson/src/main/kotlin/com/github/kittinunf/fuel/gson/FuelGson.kt
+++ b/fuel-gson/src/main/kotlin/com/github/kittinunf/fuel/gson/FuelGson.kt
@@ -22,6 +22,16 @@ inline fun <reified T : Any> Request.responseObject(noinline handler: ResponseRe
     response(gsonDeserializer(), handler)
 
 /**
+ * Asynchronously gets a response object of [T] wrapped in [Result] via [handler] using custom [Gson] instance
+ *
+ * @param gson [Gson] custom Gson deserializer instance
+ * @param handler [ResponseResultHandler<T>] the handler that is called upon success
+ * @return [CancellableRequest] request that can be cancelled
+ */
+inline fun <reified T : Any> Request.responseObject(gson: Gson, noinline handler: ResponseResultHandler<T>) =
+    response(gsonDeserializer(gson), handler)
+
+/**
  * Asynchronously gets a response object of [T] or [com.github.kittinunf.fuel.core.FuelError] via [handler]
  *
  * @param handler [Handle<T>] the handler that is called upon success
@@ -29,6 +39,17 @@ inline fun <reified T : Any> Request.responseObject(noinline handler: ResponseRe
  */
 inline fun <reified T : Any> Request.responseObject(handler: ResponseHandler<T>) =
     response(gsonDeserializer(), handler)
+
+/**
+ * Asynchronously gets a response object of [T] or [com.github.kittinunf.fuel.core.FuelError] via [handler]
+ *  using custom [Gson] instance
+ *
+ * @param gson [Gson] custom Gson deserializer instance
+ * @param handler [Handle<T>] the handler that is called upon success
+ * @return [CancellableRequest] request that can be cancelled
+ */
+inline fun <reified T : Any> Request.responseObject(gson: Gson, handler: ResponseHandler<T>) =
+    response(gsonDeserializer(gson), handler)
 
 /**
  * Synchronously get a response object of [T]
@@ -39,6 +60,15 @@ inline fun <reified T : Any> Request.responseObject() =
     response(gsonDeserializer<T>())
 
 /**
+ * Synchronously get a response object of [T]
+ *
+ * @param gson [Gson] custom Gson deserializer instance
+ * @return [Triple<Request, Response, Result<T, FuelError>>] the deserialized result
+ */
+inline fun <reified T : Any> Request.responseObject(gson: Gson) =
+    response(gsonDeserializer<T>(gson))
+
+/**
  * Generate a [ResponseDeserializable<T>] that can deserialize json of [T]
  *
  * @return [ResponseDeserializable<T>] the deserializer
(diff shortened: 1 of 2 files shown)


kittinunf:
Hey! Thanks a ton for this! Please allow me to take a look at the changes ✨ 🍻 .

fuel-gson/src/main/kotlin/com/github/kittinunf/fuel/gson/FuelGson.kt
```diff
@@ -21,6 +21,16 @@ import java.io.Reader
...
     response(gsonDeserializer(), handler)
 
+/**
+ * Asynchronously gets a response object of [T] wrapped in [Result] via [handler] using custom [Gson] instance
+ *
+ * @param gson [Gson] custom Gson deserializer instance
+ * @param handler [ResponseResultHandler<T>] the handler that is called upon success
+ * @return [CancellableRequest] request that can be cancelled
+ */
+inline fun <reified T : Any> Request.responseObject(gson: Gson, noinline handler: ResponseResultHandler<T>) =
```
kittinunf: While we don't have anything particularly to against this, this obviously also dose the job. But what do you think about what we have done in the `fuel-moshi` extension? 

Here is the link to the master version 

https://github.com/kittinunf/fuel/blob/master/fuel-moshi/src/main/kotlin/com/github/kittinunf/fuel/moshi/FuelMoshi.kt#L15

In short, in Moshi extension, we just expose the `Moshi.Builder()` directly. What do you think about it? I think it might be just *easier* and *consistent* to do the same thing?

fuel-gson/src/test/kotlin/com/github/kittinunf/fuel/FuelGsonTest.kt
```diff
@@ -30,6 +38,33 @@ private data class IssueInfo(
...
+                }
+            } else {
+                throw Error("String expected")
+            }
+    }
+}
+
+private typealias IssueTypeList = List<IssueType>
+
+
```
kittinunf: [super nits] Can you please remove an extra empty line 🙏 ?

fuel-gson/src/main/kotlin/com/github/kittinunf/fuel/gson/FuelGson.kt
```diff
@@ -30,6 +40,17 @@ inline fun <reified T : Any> Request.responseObject(noinline handler: ResponseRe
...
 
+/**
+ * Asynchronously gets a response object of [T] or [com.github.kittinunf.fuel.core.FuelError] via [handler]
+ *  using custom [Gson] instance
+ *
+ * @param gson [Gson] custom Gson deserializer instance
+ * @param handler [Handle<T>] the handler that is called upon success
+ * @return [CancellableRequest] request that can be cancelled
+ */
+inline fun <reified T : Any> Request.responseObject(gson: Gson, handler: ResponseHandler<T>) =
```
kittinunf: Also, same comment as https://github.com/kittinunf/fuel/pull/645/files#r289631684

fuel-gson/src/test/kotlin/com/github/kittinunf/fuel/FuelGsonTest.kt
lostintime: Sorry, forgot to run ktlint right before push, fixed

kittinunf: No worries ❤️ (_I am not sure why our CI didn't detect this though)

fuel-gson/src/main/kotlin/com/github/kittinunf/fuel/gson/FuelGson.kt
lostintime: Isn't `Moshi.Builder()` mutable? I'm afraid users can end messing things up, ex: you may want to decode same type from different APIs with different formats, I usually prefer to not expose unsafe APIs, but completely support you about consistency

Of course both options may be provided - shared mutable builder and explicit argument.

Also users may fallback to using `gsonDeserializer(Gson)` directly, in this case that function is not needed.

This is how my temporary solution currently looks:
```kotlin
inline fun <reified T : Any> Gson.deserializer(): ResponseDeserializable<T> =
    object : ResponseDeserializable<T> {
        override fun deserialize(reader: Reader): T? = fromJson<T>(reader, object : TypeToken<T>() {}.type)
    }

Fuel
    .get("path/to/my/service")
    .timeoutRead(timeoutMillis)
    .responseObject(gson.deserializer<List<SavedUserFeedEvent>>()) { result ->
        Either.run {
            result.fold(success = { cb(right(it)) }, failure = { cb(left(it)) })
        }
    }
```

kittinunf: That's true, I think you are right. While we are exposing the mutable builder, it provides the ease of use (with the power of users can mess things up of course). But what you said is correct about exposing the safe API is usually a better choice.

I think it depends on how much compromisation we would love to make. I think maybe exposing 2 APIs options also makes sense and giving flexibility.

As you are obviously one of the users! 😄, WDYT? In whatever way, we come to the conclusion, I will open a PR for `moshi` to standardize this.

fuel-gson/src/test/kotlin/com/github/kittinunf/fuel/FuelGsonTest.kt
iNoles: reviewdog only run when there is a fail on Travis's PR.

fuel-gson/src/main/kotlin/com/github/kittinunf/fuel/gson/FuelGson.kt
lostintime: I _vote_ for having explicit arguments for `responseObject()` functions and this will be consistent with [FuelJackson](https://github.com/kittinunf/fuel/blob/4fd0722c90d1cddfed5593fea53f502faadf7a85/fuel-jackson/src/main/kotlin/com/github/kittinunf/fuel/jackson/FuelJackson.kt) implementation.

I would strongly discourage using shared builder, but ... :) - to keep it consistent - or need to add it to `fuel-gson` also, or deprecate it from other implementations - which is a _breaking change_: so this is more a decision for project owner, my view on the project is very limited.

We can keep it as is for now and it may be added later, when somebody request it: not giving it now - is _ok_ ... but giving it now and getting it back later - will be a breaking change

SleeplessByte: I think it's good to:

- deprecate shared builder (don't remove, yet), add default builder options to both (or all integrations)
- add function call time builder argument that defaults to a **new** builder using the **default** options.

kittinunf: Let's do that then? @lostintime would you mind doing us a favor of providing both options? We will be taking care of some other extensions.

lostintime: Sure, (was a bit busy these days :)

So, just to confirm - do you want me to add a `val defaultGson = GsonBuilder()` instance and mark it as deprecated from the beginning? or just add default value to `gsonDeserializer(gson = Gson())` function, and use it everywhere needed?

kittinunf: > just add default value to gsonDeserializer(gson = Gson()) function, and use it everywhere needed?

Ah I meant this!

lostintime: I've moved Gson() instance to gsonDeserializer agrument.

I've also changed `gsonDeserializerOf` to:

```kotlin
inline fun <reified T : Any> gsonDeserializerOf(clazz: Class<T>) = gsonDeserializer<T>()
```
It seemed redundant to me and there was one ore `Gson` instance used, but I'm not sure this is right, or maybe add one more gson argument here also, with default value?

lostintime: *probably it would make sense if you use this method from Java code ...

kittinunf: Yes, that sounds great with a default value. But I think I am happy to see this in. 🎉

kittinunf:
Let's merge this.

PR #653, opened 2019-06-26
Fix content length parameter for InlineDataPart

ASKabanets:
## Description
Hi Fuel team, right now we are using your library as main networking tool and we have several multipart data request for uploading images. 
What I have found, that in your InlineDataPart you are using `content.length.toLong()` which is not right. For example, you can have emoji in your content and measuing size of the DataPart by string length doesn't seem proper.
So instead I changed measuring size of the InlineDataPart to `content.toByteArray().size`

I didn't find a issue related to this problem and created pull request right away. I hope you will find and useful and merge into the master, because its really crucial for us.

## Type of change

Check all that apply

## How Has This Been Tested?

Tested it with several strings with emoji. 

## Checklist:

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/DataPart.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/core/DataPart.kt
@@ -100,7 +100,7 @@ data class InlineDataPart(
     override val contentDisposition: String = "form-data; name=\"$name\"${if (filename != null) "; filename=\"$filename\"" else "" }"
 ) : DataPart() {
     override fun inputStream() = content.byteInputStream()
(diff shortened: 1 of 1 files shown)


SleeplessByte:
> `content.toByteArray().size`

Does this take in account the target encoding?

ASKabanets:
@SleeplessByte Its a good question, I updated code so now it takes in account encoding

SleeplessByte:
> What I have found, that in your InlineDataPart you are using content.length.toLong() which is not right. For example, you can have emoji in your content and measuing size of the DataPart by string length doesn't seem proper.

It would be great if you could add _one_ test case that shows this (have it have emojis and test the length, the old code should be incorrect and the new code should be correct). 

Would you be able to do that?

PR #676, opened 2019-08-14
[fuel-core] Use StreamingMode only in UploadRequest

caiooliveiraeti:
When StreamingMode is used cannot get response data from status code 401. https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html#setFixedLengthStreamingMode-long-

In Fuel version 1.X.X you can get the response data, that's because StreamingMode was only used for UploadRequest.

## Type of change

Check all that apply

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
@@ -9,6 +9,7 @@ import com.github.kittinunf.fuel.core.Method
 import com.github.kittinunf.fuel.core.Request
 import com.github.kittinunf.fuel.core.Response
 import com.github.kittinunf.fuel.core.requests.DefaultBody
+import com.github.kittinunf.fuel.core.requests.UploadRequest
 import com.github.kittinunf.fuel.core.requests.isCancelled
 import com.github.kittinunf.fuel.util.ProgressInputStream
 import com.github.kittinunf.fuel.util.ProgressOutputStream
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
@@ -229,12 +229,14 @@ class HttpClient(
         }
 
         val contentLength = body.length
-        if (contentLength != null && contentLength != -1L) {
-            // The content has a known length, so no need to chunk
-            connection.setFixedLengthStreamingMode(contentLength.toLong())
-        } else {
-            // The content doesn't have a known length, so turn it into chunked
-            connection.setChunkedStreamingMode(4096)
+        if (request is UploadRequest) {
+            if (contentLength != null && contentLength != -1L) {
+                // The content has a known length, so no need to chunk
+                connection.setFixedLengthStreamingMode(contentLength.toLong())
+            } else {
+                // The content doesn't have a known length, so turn it into chunked
+                connection.setChunkedStreamingMode(4096)
+            }
         }
 
         val noProgressHandler = request.executionOptions.requestProgress.isNotSet()


SleeplessByte:
I'm able to get response data just fine when streaming mode is on? Do you have a reproducible example? If you absolute _must_ have this, instead let's add a flag that allows you to turn off setting streaming mode, instead of changing everyone's behaviour.

caiooliveiraeti:
Hello @SleeplessByte.

I am trying to migrate from version 1.X.X to 2.2.0, my test cases have broken because I cannot get response data when status code is 401. This is because HTTPURLConnection disconnect before reading InputStream: 

> ["When output streaming is enabled, authentication and redirection cannot be handled automatically. A HttpRetryException will be thrown when reading the response if authentication or redirection are required. This exception can be queried for the details of the error."](https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html#setFixedLengthStreamingMode-long-)

**Fuel 1.15.1**
<-- 401 (http://localhost:7000/test-error-streaming-mode)

> Response : Unauthorized
> Length : 18
> **Body : ({“result”:”error”})**
> Headers : (4)
> Server : [Custom]
> Content-Length : [18]
> Date : [Wed, 14 Aug 2019 14:21:37 GMT]
> Content-Type : [application/json;charset=utf-8]

**Fuel 2.2.0**

> <-- 401 http://localhost:7000/test-error-streaming-mode
> Response : Unauthorized
> Length : 18
> **Body : (18 bytes of application/json;charset=utf-8)**
> Headers : (4)
> Date : Wed, 14 Aug 2019 14:08:49 GMT
> Server : Custom
> Content-Type : application/json;charset=utf-8
> Content-Length : 18

**Fuel 2.2.0 with change**

> <-- 401 http://localhost:7000/test-error-streaming-mode
> Response : Unauthorized
> Length : 18
> **Body : {“result”:”error”}**
> Headers : (4)
> Date : Wed, 14 Aug 2019 14:27:01 GMT
> Server : Custom
> Content-Type : application/json;charset=utf-8
> Content-Length : 18

SleeplessByte:
But this sounds like the solution is _capturing `HttpRetryException`_ (probably in `Validator/ValidateResponseCode` middleware) and not turning off streaming for everyone.

caiooliveiraeti:
When throwing the `HttpRetryException` the [connection is closed](https://github.com/JetBrains/jdk8u_jdk/blob/041d7dd0d734d56458c2b6f820ad7bba5b5ed087/src/share/classes/sun/net/www/protocol/http/HttpURLConnection.java#L1689). That way we won't be able to get the data from `getInputStream()` or `getErrorStream()`.

To get response data in status code 401 cases, you must disable the request Streaming Mode.

What is the best solution to do this?

caiooliveiraeti:
### Streaming mode

The [HttpURLConnection](http://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html) will by default buffer the entire request body before actually sending it, regardless of whether you've set a fixed content length yourself using `connection.setRequestProperty("Content-Length", contentLength);`. This may cause `OutOfMemoryExceptions` whenever you concurrently send large POST requests (e.g. uploading files). To avoid this, you would like to set the 
[HttpURLConnection#setFixedLengthStreamingMode()](https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html#setFixedLengthStreamingMode-int-).

> httpConnection.setFixedLengthStreamingMode(contentLength);

But if the content length is really not known beforehand, then you can make use of chunked streaming mode by setting the [HttpURLConnection#setChunkedStreamingMode()](https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html#setChunkedStreamingMode-int-) accordingly. This will set the HTTP [Transfer-Encoding](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.41) header to `chunked` which will force the request body being sent in chunks. The below example will send the body in chunks of 1KB.

> httpConnection.setChunkedStreamingMode(1024);

By [How to use java.net.URLConnection to fire and handle HTTP requests](https://stackoverflow.com/questions/2793150/how-to-use-java-net-urlconnection-to-fire-and-handle-http-requests/2793153#2793153)

SleeplessByte:
I'll investigate this and get back to you. Thanks for reporting!

PR #700, opened 2019-11-20
:art: Add EmptyDeserializer and corresponding extensions methods

xJoeWoo:
## Description

Recently I used `Request.download()` and `Request.awaitString()` functions to download a large file. However `awaitString()` function threw an exception about memory allocation too large.

Existing functions like `Request.responseString()`, `Request.response()`, `Request.await(Deserializable)` which used either `ByteArrayDeserializer` or `StringDeserializer` will deserialize result then may cause the same problem.

Then I noticed the  [document](https://fuel.gitbook.io/documentation/core/fuel#download-response-to-output-file-or-outputstream) metioned an `EmptyDeserializer` can be used to both download the file and avoid getting result. But `EmptyDeserializer` was not found in library.

This PR added a `EmptyDeserializer` and some extension functions(in core, coroutine, reactor, rx modules) named `***Ignored()`. They can be used with `download()` or somewhere users don't want to retrieve the response result.

## Type of change

Check all that apply

## How Has This Been Tested?

Added corresponding unit tests and had them passed.

## Checklist:

### DIFF
--- a/fuel-reactor/src/main/kotlin/com/github/kittinunf/fuel/reactor/Reactor.kt
+++ b/fuel-reactor/src/main/kotlin/com/github/kittinunf/fuel/reactor/Reactor.kt
@@ -99,4 +99,4 @@ fun <T : Any> Request.monoResultObject(mapper: Deserializable<T>): Mono<Result<T
 /**
  * Get a complete signal(success with [Unit]) via a [MonoSink.success], or any [FuelError] via [MonoSink.error]
  */
-fun Request.monoIgnored(): Mono<Unit> = monoResultFold(EmptyDeserializer)
+fun Request.monoUnit(): Mono<Unit> = monoResultFold(EmptyDeserializer)
--- a/fuel-reactor/src/main/kotlin/com/github/kittinunf/fuel/reactor/Reactor.kt
+++ b/fuel-reactor/src/main/kotlin/com/github/kittinunf/fuel/reactor/Reactor.kt
@@ -1,10 +1,14 @@
 package com.github.kittinunf.fuel.reactor
 
-import com.github.kittinunf.fuel.core.*
+import com.github.kittinunf.fuel.core.Deserializable
+import com.github.kittinunf.fuel.core.FuelError
+import com.github.kittinunf.fuel.core.Request
+import com.github.kittinunf.fuel.core.Response
 import com.github.kittinunf.fuel.core.deserializers.ByteArrayDeserializer
 import com.github.kittinunf.fuel.core.deserializers.EmptyDeserializer
 import com.github.kittinunf.fuel.core.deserializers.StringDeserializer
 import com.github.kittinunf.fuel.core.requests.CancellableRequest
+import com.github.kittinunf.fuel.core.response
 import com.github.kittinunf.result.Result
 import reactor.core.publisher.Mono
 import reactor.core.publisher.MonoSink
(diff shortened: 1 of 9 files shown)


kittinunf:
Thank you so much for this! I think this is a good addition to the library!

I wonder whether we should stick with the naming `Unit` instead of `Ignore` a self-document for the return type? Like `awaitUnit`, `rxUnit` instead

xJoeWoo:
I think `**Unit` is better as well, in fact, I used this naming before commit. 😋 I'll make another commit to change naming to `Unit` suffix later.

fuel-rxjava/src/main/kotlin/com/github/kittinunf/fuel/rx/RxFuel.kt
```diff
@@ -224,6 +225,11 @@ fun <T : Any> Request.rxObjectPair(deserializable: Deserializable<T>) = rxResult
  */
 fun <T : Any> Request.rxObjectTriple(deserializable: Deserializable<T>) = rxResultTriple(deserializable)
 
+/**
+ * Returns a reactive stream for a [Single] value with only complete(success with [Unit]) or error signal
+ */
+fun Request.rxUnit(): Single<Unit> = rxResponseSingle(EmptyDeserializer)
```
kittinunf: 🤔 Since we are no longer care about the emitting data? How about turning this into a `Completable`, I am not really sure whether the `Single` still fits the bill here anymore. WDYT?

xJoeWoo: Oh.. You're right, I can't find `onSuccess` function without passing an argument in Rx, however that's the way to complete an stream in Reactor which I'm more familiar with. `Completable` should be the chosen one.

fuel-reactor/src/main/kotlin/com/github/kittinunf/fuel/reactor/Reactor.kt
```diff
@@ -1,13 +1,10 @@
 package com.github.kittinunf.fuel.reactor
 
-import com.github.kittinunf.fuel.core.Deserializable
-import com.github.kittinunf.fuel.core.FuelError
-import com.github.kittinunf.fuel.core.Request
-import com.github.kittinunf.fuel.core.Response
+import com.github.kittinunf.fuel.core.*
```
kittinunf: [nits] Can you please not `*` this?

fuel-rxjava/src/main/kotlin/com/github/kittinunf/fuel/rx/RxFuel.kt
```diff
@@ -1,15 +1,12 @@
-
 package com.github.kittinunf.fuel.rx
 
-import com.github.kittinunf.fuel.core.Deserializable
-import com.github.kittinunf.fuel.core.FuelError
-import com.github.kittinunf.fuel.core.Request
-import com.github.kittinunf.fuel.core.Response
+import com.github.kittinunf.fuel.core.*
```
kittinunf: Here too 🙇‍♀

kittinunf:
I think this looks good now, and I am happy to merge this!! 👍 Can you please fix the import?

xJoeWoo:
Ah.. IntelliJ auto optimized imports with wildcard when I commit. I restored them from older commits.

Thanks for accepting my PR, very glad that I can help to make such excellent library better. 🥰

kittinunf:
Lovely, 💖💖💖 thanks so much for your contribution!

PR #701, opened 2019-11-23
setFixedLengthStreamingMode Fix for API < 19

rumboalla:
## Description

setFixedLengthStreamingMode(Long) requires at least Android API 19. If that method is not found, call the old setFixedLengthStreamingMode(Int).

## Type of change

Check all that apply

## How Has This Been Tested?

Tested this with my app and API 15.

## Checklist:


iNoles:
That is great for personal objects, but not for business.

SleeplessByte:
Keeping minsdk low because you have < 1% of your users on annold phone that *no longer receives security updates* is a bad business practice.

I see absolutely no value supporting an SDK below 21, let alone 19.

kittinunf:
I kinda agree with @SleeplessByte, I think we should support > 21 ... Shall we bump the minSDK instead of this?

@rumboalla Do you have a strong reason to support < 21?

kittinunf:
Let's update API to level 21, shall we?

PR #755, opened 2020-06-10
Fix content-length header not being sent when messages have empty body

### LINKED ISSUE #748: The Content-Length header is not included in requests
# Bug Report

## Description
The `Content-Length` header is not included in requests. I did not find out how to prevent this. Is it intentional or actually a bug? I did not find relevant documentation on this behavior of Fuel .

This keeps Fuel ⛽ from being used for APIs that respond with [HTTP 411](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/411) is the field is not present. This is the case for example with the [Google Play Developer API](https://developers.google.com/android-publisher/api-ref/).

## To Reproduce
Code to reproduce the behavior:

```Kotlin
import com.github.kittinunf.fuel.Fuel
import com.github.kittinunf.fuel.core.Headers

println(Fuel
        .post("https://httpbin.org/anything")
        .header(Headers.CONTENT_LENGTH, "0")
        .responseString()
        .third
        .component1())
```

Output:
```JSON
{
  "args": {}, 
  "data": "", 
  "files": {}, 
  "form": {}, 
  "headers": {
    "Accept": "text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2", 
    "Content-Type": "application/x-www-form-urlencoded", 
    "Host": "httpbin.org", 
    "User-Agent": "Java/13.0.1", 
    "X-Amzn-Trace-Id": "Root=1-XXXXXXX-XXXXXXXXXXXXXXXXX"
  }, 
  "json": null, 
  "method": "POST", 
  "origin": "--------", 
  "url": "https://httpbin.org/anything"
}

```

### Development Machine

 - OS: Windows 10
 - Fuel version: 2.2.2
 - Kotlin version: 1.3.72 @ Java/13.0.1

felixdivo:
## Description

It happened because streaming mode was used in these cases. See #748 The fix can be seen as a fast path: It would probably not be required to have the `isEmpty()` check in the first place for correctness alone.

Fixes #748

## Type of change

Check all that apply

## How Has This Been Tested?

Manually, and a unit/regression test was added.

## Checklist:

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/HttpClient.kt
@@ -234,7 +234,11 @@ class HttpClient(
 
     private fun setBodyIfDoOutput(connection: HttpURLConnection, request: Request) {
         val body = request.body
-        if (!connection.doOutput || body.isEmpty()) {
+        if (!connection.doOutput) {
+            return
+        }
+        if (body.isEmpty()) {
+            connection.setFixedLengthStreamingMode(0L)
             return
         }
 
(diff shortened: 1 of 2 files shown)


felixdivo:
cc @SleeplessByte

SleeplessByte:
This is a bug fix, not a breaking change :)

PR #764, opened 2020-07-30
Update Kotlin (1.4.0) from 3x branch

iNoles:
Prepare to update to Kotlin 1.4.0 once it is stable.


fuel-singleton/src/main/kotlin/fuel/HttpLoaderFactory.kt
```diff
@@ -1,7 +1,8 @@
 // Inspired by https://github.com/coil-kt/coil/blob/master/coil-default/src/main/java/coil/ImageLoaderFactory.kt
+
 package fuel
 
-interface HttpLoaderFactory {
+fun interface HttpLoaderFactory {
```
kittinunf, line 5: 👍 1.4 new stuff right?

iNoles: Yup

kittinunf:
Do we need to do something about our Lint Code? 🤔

iNoles:
> Do we need to do something about our Lint Code? 🤔
> 
> ```
> [/github/workspace/fuel-singleton/src/main/kotlin/fuel/HttpLoaderFactory.kt:5:4: Not a valid Kotlin file (5:4 expecting function name or receiver type) (cannot be auto-corrected)]
> ```

that is a bug from ktlint. https://github.com/pinterest/ktlint/issues/726

kittinunf:
Thanks @iNoles 👏

Issue #776, opened 2020-10-11
Assertions in Jackson extensions tests never fail


detouched:
Some tests in `FuelJacksonTest` class have assertions that will never fail the test.

The problem is that assertions are executed in the asynchronous response handler. Since the request result is not awaited, they will error out on a non-main thread, so test engine will not know about the failures.

## To Reproduce
1. Put any obviously false assertion into, for example, [`jacksonTestResponseObject()`](https://github.com/kittinunf/fuel/blob/572d907/fuel-jackson/src/test/kotlin/com/github/kittinunf/fuel/FuelJacksonTest.kt#L41-L43)
2. Run the test
3. The test will pass though should've failed

## Potential solution
Probably the easiest would be to append `.get()` for every asynchronous call which is inherited from `Future<Response>`. This will let the error bubble  up. This solution, however, depends on fixing #775 because some asynchronous extension functions don't expose `CancellableRequest` object.

detouched:
In fact, 9 out of 15 tests actually fail now if I just add proper awaiting on the result in them. Hopefully they're just out of date, not catching real bugs :)

iNoles:
Should this be closed?

SleeplessByte:
Yeah. 

The PR had Fixes a and b. 

GitHub only understands 

Fixes a
Fixes b

detouched:
Oops, TIL my bad, sorry.

PR #777, opened 2020-10-11
Fix Jackson extension functions and their tests

### LINKED ISSUE #775: Jackson extensions: responseObject() should return CancellableRequest object
Some [`responseObject(...)`](https://github.com/kittinunf/fuel/blob/63c5f4c/fuel-jackson/src/main/kotlin/com/github/kittinunf/fuel/jackson/FuelJackson.kt#L20-L26) extension methods for Jackson don't expose result of `CancellableRequest` type, which makes them inconsistent with other helper methods and limits their use cases for no reason (e.g. no way to `cancel()` or `join()` request when these shorthand methods used).

Specifically, shorthand methods that accept the triple of _request_, _response_ and _result_ have `Unit` return type.

detouched:
This fixes the tests of Jackson extension functions for asynchronous requests (#776). These tests were silently failing for a long time because of the assertions in the callbacks which are async and not visible to the test engine. As a result, the tests became out of date, so aside from ensuring the assertions are now able to fail the tests this PR also contains a decent amount of changes in the setup procedures as well as assertions themselves.

Assertions are now ensured to take an effect on the test outcome by explicitly awaiting on the `CancellableRequest` object which is returned by the extension functions using the `get()` method inherited from `Future<Response>`.
This change also fixes two extension functions that did not return `CancellableRequest` object for no reason (#775).

I also fixed code formatting around the lines I touched according to what seems to be the most common in this project.

Fixes #775 and #776.

## Type of change

## How Has This Been Tested?

I did not include any new tests for the change in the production code (#775) 
because existing tests do verify that change after the fix of #776.

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas **(n/a)**
- [ ] Inspect the bytecode viewer, including reasoning why **(n/a)**

### DIFF
--- a/fuel-jackson/src/test/kotlin/com/github/kittinunf/fuel/FuelJacksonTest.kt
+++ b/fuel-jackson/src/test/kotlin/com/github/kittinunf/fuel/FuelJacksonTest.kt
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.databind.PropertyNamingStrategy
 import com.fasterxml.jackson.module.kotlin.registerKotlinModule
 import com.github.kittinunf.fuel.core.FuelError
+import com.github.kittinunf.fuel.core.HttpException
 import com.github.kittinunf.fuel.core.Request
 import com.github.kittinunf.fuel.core.Response
 import com.github.kittinunf.fuel.core.ResponseHandler
@@ -20,6 +21,7 @@ import org.hamcrest.CoreMatchers.notNullValue
 import org.hamcrest.CoreMatchers.nullValue
 import org.junit.Assert.assertNotEquals
 import org.junit.Assert.assertThat
+import org.junit.Assert.fail
 import org.junit.Test
 import org.mockserver.matchers.Times
 import java.net.HttpURLConnection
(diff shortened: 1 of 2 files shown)


fuel-jackson/src/test/kotlin/com/github/kittinunf/fuel/FuelJacksonTest.kt
```diff
@@ -38,25 +40,33 @@ class FuelJacksonTest : MockHttpTestCase() {
...
-                assertThat(result.component1(), instanceOf(HttpBinUserAgentModel::class.java))
-                assertThat(result.component1()?.userAgent, not(""))
-                assertThat(result.component2(), instanceOf(FuelError::class.java))
+                assertThat(result, instanceOf(Result.Success::class.java))
+                with(result as Result.Success) {
+                    assertThat(value, instanceOf(HttpBinUserAgentModel::class.java))
+                    assertThat(value.userAgent, not(""))
+                }
             }
+            .get()
```
detouched, line 49: This ensures that assertions in the asynchronous callback are able to fail the tests.

SleeplessByte (approved):
Looks like this fixes the issue!

kittinunf:
This is amazinggg! Thanks!

kittinunf (approved):

PR #785, opened 2020-11-24
Fix: PATCH on HTTPS endpoint

### LINKED ISSUE #784: PATCH on HTTPS endpoint doesn't work
# Bug Report

## Description
Using PATCH method on a HTTPS endpoint doesn't work even when setting `FuelManager.instance.forceMethods = true`

## To Reproduce
```Kotlin
    FuelManager.instance.forceMethods = true
    val (_, _, result) = Fuel.patch("https://httpbin.org/patch")
            .also { println(it) }
            .response()
    if (result is Failure) {
        println(result.error) // HTTP Exception 405 METHOD NOT ALLOWED
    }
```

## Expected behavior
When setting `FuelManager.instance.forceMethods = true`, the HTTP method should be PATCH and not POST with header `X-HTTP-Method-Override`

cpuydebois:
## Description
The `forceMethod` extension used to set the correct HTTP method when using PATCH uses reflexion to force the value of the `method` field in the `HttpURLConnection` class.

In the case of HTTPS, the class hierarchy of the connection class is more complex, and the `forceMethod` extension doesn't go back enough in the parent classes to find the correct `method` field to set.

This PR fixes the `forceMethod` extension.

Fixes #784 

## Type of change

## Checklist:

### DIFF
--- a/fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/extensions/ForceMethod.kt
+++ b/fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/extensions/ForceMethod.kt
@@ -20,16 +20,20 @@ fun HttpURLConnection.forceMethod(method: Method) {
             } catch (ex: NoSuchFieldException) {
                 // ignore
             }
-            (arrayOf(this.javaClass.superclass, this.javaClass)).forEach {
-                try {
-                    it.getDeclaredField("method").apply {
-                        this.isAccessible = true
-                    }.set(this, method.value)
-                } catch (ex: NoSuchFieldException) {
-                    // ignore
-                }
-            }
+            this.forceMethod(this.javaClass, method)
         }
         else -> this.requestMethod = method.value
+    }
+}
+
+private fun HttpURLConnection.forceMethod(clazz: Class<in HttpURLConnection>, method: Method) {
+    try {
+        clazz.getDeclaredField("method").apply {
+            this.isAccessible = true
+        }.set(this, method.value)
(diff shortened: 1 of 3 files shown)


fuel/src/main/kotlin/com/github/kittinunf/fuel/toolbox/extensions/ForceMethod.kt
```diff
@@ -20,16 +20,20 @@ fun HttpURLConnection.forceMethod(method: Method) {
...
+}
+
+private fun HttpURLConnection.forceMethod(clazz: Class<in HttpURLConnection>, method: Method) {
+    try {
+        clazz.getDeclaredField("method").apply {
+            this.isAccessible = true
+        }.set(this, method.value)
+    } catch (ex: NoSuchFieldException) {
+        if (HttpURLConnection::class.java.isAssignableFrom(clazz.superclass)) {
+            this.forceMethod(clazz.superclass, method)
```
SleeplessByte, line 36: This is what the fix boils down to, right?

cpuydebois: That's right.

More specifically, by replacing the iteration : 

```kotlin
(arrayOf(this.javaClass.superclass, this.javaClass)).forEach {
```

with a new recursive method which traverses the hierarchy of parent classes until it finds the correct one.

SleeplessByte (approved):

SleeplessByte:
Makes sense to me!

kittinunf (approved):
GO!

PR #793, opened 2021-02-22
[3.x] Add synchronous blocking style interface

kittinunf:
### What's in this PR?

This is a PR to introduce the core that exposes the `Call` so it opens the way for converting to any other type later.


fuel-base/src/main/kotlin/fuel/HttpLoader.kt
```diff
@@ -22,6 +22,20 @@ public interface HttpLoader {
...
     public suspend fun method(request: Request): Response
 
+    /**
+     * Synchronously load the [request]'s data and suspend until the operation is complete. Return the loaded [Response]. This is the counterpart of the suspend version above
+     *
+     * @param request The request to execute synchronously.
+     * @return The [Response] result.
+     */
+    public fun getBlocking(request: Request): Response
+    public fun postBlocking(request: Request): Response = TODO("Not implemented yet")
```
kittinunf, line 32: I will continue on this after the direction of this PR is agreed upon first.

fuel-samples/progress/build.gradle.kts
```diff
@@ -3,5 +3,5 @@ application {
 }
 
 dependencies {
-    implementation(project(":fuel-singleton"))
```
kittinunf, line 6: We don't need `fuel-singleton` in this sample so make it so.

fuel-base/src/test/kotlin/fuel/RealHttpLoaderTest.kt
```diff
@@ -266,7 +266,7 @@ internal class RealHttpLoaderTest {
             mockWebServer.takeRequest()
         }
 
-        assertEquals("GET", request1.method)
+        assertEquals("HEAD", request1.method)
```
kittinunf, line 269: This supposes to be "HEAD" right?

fuel-base/src/main/kotlin/fuel/HttpLoader.kt
```diff
@@ -22,6 +22,20 @@ public interface HttpLoader {
...
+     * @return The [Response] result.
+     */
+    public fun getCall(request: Request): Call
+    public fun postCall(request: Request): Call
+    public fun putCall(request: Request): Call
+    public fun patchCall(request: Request): Call
+    public fun deleteCall(request: Request): Call
+    public fun headCall(request: Request): Call
+    public fun methodCall(request: Request): Call
+
```
SleeplessByte, line 38: Are these shortcuts for `method()`?

iNoles: Well, it supposed to be the custom http methods. It can be used as get, post and others too.

SleeplessByte: Yeah, as long as there is a way to do non-common HTTP methods (via `method` for example), it's all good.

iNoles:
I don't think synchronous would be smartest thing to do for the HTTP Client. 

By looking at https://github.com/Kotlin/kotlinx.coroutines, it already have support for Reactive-Related Libraries, CompletionStage (JDK8), Guava and Task (Google Play Services). Oh course, it is just my opinions.

fuel-base/src/main/kotlin/fuel/Extensions.kt
```diff
@@ -31,6 +31,14 @@ internal suspend fun Call.await(): Response = suspendCancellableCoroutine { cont
     })
 }
 
+/**
+ * Call response extension that allows to execute the Call with validator
+ *
+ * @return Response of response with validator by default it uses [Response.isSuccessful] to determine whether response
+ * is successful or not, otherwise throw exception
+ */
+public fun Call.response(validator: Response.() -> Response = Response::validate): Response = execute().validator()
```
iNoles, line 40: Is there is a reason for validator parameters?

kittinunf: I think we should give some room for the client to customize the way we validate the response and throw exception.

fuel-base/src/main/kotlin/fuel/RealHttpLoader.kt
```diff
@@ -6,48 +6,108 @@ import okhttp3.Call
 import okhttp3.Request.Builder
 import okhttp3.Response
 
+private enum class Method {
```
iNoles, line 9: I believe we should drop this class for strings. See reasons in another comments

fuel-base/src/main/kotlin/fuel/RealHttpLoader.kt
```diff
@@ -6,48 +6,108 @@ import okhttp3.Call
...
         val requestBuilder = Builder().headers(request.headers).method(method, request.requestBody)
-        return fetcher.fetch(request.data, requestBuilder)
+        return fetcher.fetch(request.data, requestBuilder).await().validate()
+    }
+    //endregion
+}
+
+private fun createRequestBuilder(request: Request, method: Method): Builder {
+    return Builder().headers(request.headers).apply {
+        if (method == Method.GET || method == Method.HEAD) {
```
iNoles, line 100: We should use https://github.com/square/okhttp/blob/master/okhttp/src/main/kotlin/okhttp3/internal/http/HttpMethod.kt#L26 for better checks.

fuel-base/src/main/kotlin/fuel/RealHttpLoader.kt
kittinunf: Ok, will update this

kittinunf: Done!

fuel-base/src/main/kotlin/fuel/RealHttpLoader.kt
kittinunf: Done

iNoles (approved):
Much better

PR #798, opened 2021-03-06
[3.x] Add more tests and expose interface for singleton

kittinunf:
### What's in this PR?

Add more tests and expose the interface for the client of singleton to use.

### DIFF
--- a/fuel-singleton/build.gradle.kts
+++ b/fuel-singleton/build.gradle.kts
@@ -1,5 +1,10 @@
 apply(plugin = "com.vanniktech.maven.publish")
 
+val okhttpVersion: String by extra
+
 dependencies {
     api(project(":fuel-base"))
+
+    testImplementation("junit:junit:4.13")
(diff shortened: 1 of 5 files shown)


iNoles:
Oh, I didn't added test for this folder before.

kittinunf:
But do you know it says that we don't have tests covered? Since we are adding tests for the interface in this PR?

iNoles:
> But do you know it says that we don't have tests covered? Since we are adding tests for the interface in this PR?

Suspend one. Well, I wouldn't worried about it.

kittinunf:
Yeah, but I have added here, right? 

https://github.com/kittinunf/fuel/pull/798/files#diff-da6ed83027be159e12093e620ff4f9447224d1608222b97dcdda9b1b4a4d4631R29-R50

iNoles:
Hmm, don't brother to add suspend test on this one. How about we closed this one to merge with my own?

kittinunf:
Ok close this PR then.

PR #820, opened 2021-11-28
📝 jcenter() is at end-of-life, recommend mavenCentral() instead

xjcl:
JCenter is at end-of-life and will be made read only on March 31st, 2021. To prevent developers from inserting a `jcenter()` declaration in their `build.gradle`, I changed the code snippet in the documentation to `mavenCentral()`. MC seems to host artifacts for >=3.2.2 and JC for <3.2.2. I also removed the JC icon at the beginning as it was not working anyway.

For more information, check this StackOverflow question: https://stackoverflow.com/questions/66400264/jcenter-is-at-end-of-life-android-lint-warning-what-is-the-replacement

I am new to this project so I hope I was useful.


kittinunf (approved):
Cool!

PR #848, opened 2023-02-28
Additional Kotlin Improvements 

iNoles:
1) Upgrade Kotlin to 1.8.10
2) Implement Parameters


build.gradle.kts
```diff
@@ -1,11 +1,9 @@
 plugins {
-    kotlin("multiplatform") version "1.7.20" apply false
-    id("org.jetbrains.kotlinx.kover") version "0.6.1" apply false
+    kotlin("multiplatform") version "1.8.10" apply false
+    id("org.jetbrains.kotlinx.kover") version "0.7.0-Alpha" apply false
```
kittinunf, line 3: Shall we use kover alpha? or it is not compatible with new Kotlin version?

iNoles: well, it is the latest version for Kover.

kittinunf: yeah, what I meant was that maybe we should stop using them until they get a bit more mature.

fuel/src/commonMain/kotlin/fuel/Request.kt
```diff
@@ -1,7 +1,10 @@
 package fuel
 
+public typealias Parameters = List<Pair<String, String>>
```
kittinunf, line 3: I'm wondering whether we should do `Pair<String, Any?>` for this.

build.gradle.kts
iNoles: Oh Okay, I see what you mean. Jacoco is not that multiplatform friendly either.

kittinunf:
Looks good for me sir! Just wondering whether we should support something rather than `Pair<String, String>`

kittinunf (approved):
🎆

PR #853, opened 2023-04-06
Fix Maven again

iNoles:
I have noticed that three JVM folders are not being uploaded to the Maven repo. I will be checking the snapshot once it is finished.

### DIFF
--- a/fuel-forge-jvm/build.gradle.kts
+++ b/fuel-forge-jvm/build.gradle.kts
@@ -7,6 +7,15 @@ kotlin {
     explicitApi()
 }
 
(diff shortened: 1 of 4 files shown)


kittinunf:
What do you think about adding this into `publication.gradle.kts` file?

```
publications {
        if (project.name.substringAfterLast("-") == "jvm") {
            create<MavenPublication>("maven") {
                from(components["java"])
            }
        }
    }
```

iNoles:
I think that would be much better.

kittinunf (approved):

PR #855, opened 2023-04-11
[Publication] Add sources jar for publication

kittinunf:
### What's in this PR? 

This PR updates the publication file a bit by adding sourcesJar into the publication.


kittinunf:
This should be replaced by #905

Issue #857, opened 2023-04-16
No Option for headers in Fuel.Get for 3.X.X


geofferyj:

geofferyj:
There is also no way to get error messages

iNoles:
> There is also no way to get error messages

the library will throw Exceptions. You can use simple try/catch to get the error messages.

geofferyj:
Great thanks.

PR #858, opened 2023-04-17
Add Headers on Get

iNoles:
This was one suggestion from #857


kittinunf (approved):
Nice one!

kittinunf:
Can we add unit test quickly for this too? We also can do it in a separate PR.

iNoles:
> Can we add unit test quickly for this too? We also can do it in a separate PR.

Sure, I would have to do it another day. I don't mind if you do it if you have time.

Issue #859, opened 2023-04-18
Make it easier to mock Fuel behavior in v3


dalewking:
For testing purposes one needs to set Fuel up to mock the behavior. You do provide:
```
    public fun setHttpLoader(loader: HttpLoader)
    public fun setHttpLoader(factory: HttpLoaderFactory)
```

But those are kind of unusable, since they are all about HttpLoader, which is an actual/expect class and is final so it is not possible to actually create a custom version of HttpLoader.

HttpLoader needs to be an interface. It would be cleaner you instead created a global val defaultHttpLoader that was expect/actual rather than trying to have an expect actual class.

iNoles:
Should this to be closed?

PR #861, opened 2023-04-19
Each Platforms override HttpLoader's methods

iNoles:
Thanks for the suggestions from #859.  Did this meet your approval, @dalewking


dalewking:
That is definitely a step in the right direction.

I see another improvement probably not directly related to my request, but this change makes it easier. I would ask why Fuel class is using expect/actual? Some of the methods don't vary at all across platforms. The only platform specific thing in there is a line like this:
```
        val loader = httpLoaderFactory?.newHttpLoader() ?: AppleHttpLoader()
```
But you already the platform specific mapping in FuelBuilder. It would seem to me that you could remove the expect actual from the Fuel class and that line could just be:

```
        val loader = httpLoaderFactory?.newHttpLoader() ?: FuelBuilder().build()
```

fuel/src/jsMain/kotlin/fuel/Fuel.kt
```diff
@@ -21,7 +21,7 @@ public actual object Fuel {
         httpLoader?.let { return it }
 
         // Create a new HttpLoader.
-        val loader = httpLoaderFactory?.newHttpLoader() ?: HttpLoader()
+        val loader = httpLoaderFactory?.newHttpLoader() ?: JSHttpLoader()
```
dalewking, line 24: Why is Fuel class still using expect/actual? Some of the methods don't vary at all across platforms. The only platform specific thing in here the lines like this one.

But you already have the platform specific mapping in FuelBuilder. It would seem to me that you could remove the expect/actual from the Fuel class and that line could just be:

```
        val loader = httpLoaderFactory?.newHttpLoader() ?: FuelBuilder().build()
```

dalewking:
Thinking about it shouldn't FuelBuilder just implement HttpLoaderFactory so you could just pass your your configured FuelBuilder to Fuel.setHttpLoader(factory)?

fuel/src/jsMain/kotlin/fuel/Fuel.kt
iNoles: Done

dalewking (approved):

kittinunf (approved):
I also like this at well! Not a big fan of actual/expect class if we can avoid this.

iNoles:
> I also like this at well! Not a big fan of actual/expect class if we can avoid this.

Trying to avoid it with the old GC for Kotlin Native for new memory management.

PR #863, opened 2023-04-27
Customized HttpResponse Per Platforms 

iNoles:
I am putting HttpResponse into expect/actual class because I knew an interface will be extremely messy for this. It will be very easier to customize it on each platform. I knew the string will be extremely inefficient for large files, so I wanted to give developers better control to handle data how they wanted. 

Breaking Change: body in HttpResponse for JVM is ResponseBody.


kittinunf:
Looks okay to me! Wondering whether we should still have a common field for something that share across the whole different platforms? e.g. statusCode, possibly more in the future.

iNoles:
> Looks okay to me! Wondering whether we should still have a common field for something that share across the whole different platforms? e.g. statusCode, possibly more in the future.

I don't think that is possible for now.

kittinunf (approved):

dalewking:
This essentially breaks fuel for use in Kotlin multiplatform. You cannot access statusCode or body in common code

PR #864, opened 2023-05-05
Upgrade Kotlin, Gradle, OkHTTP and KotlinX.

iNoles:
1) Upgrade Kotlin to 1.8.21
2) Upgrade Gradle to 8.0.2
3) Upgrade OKHTTP with Async methods
4) Upgrade Kotlinx coroutines to 1.7.0


kittinunf (approved):
👏

build.gradle.kts
```diff
@@ -1,5 +1,5 @@
 plugins {
-    kotlin("multiplatform") version "1.8.10" apply false
+    kotlin("multiplatform") version "1.8.21" apply false
```
kittinunf, line 2: Probably not related to the change in this PR, do you think we can move this version into toml file instead?

iNoles: I can do that with the next PR. Thanks to Gradle 8.0 for a better version catalog.

PR #865, opened 2023-05-07
Fix Maven Publications' Signing

iNoles:
This will fix Maven Signing with proper tasks by depending on the correct sign class.


kittinunf (approved):
Thank you so much! Let's goo!

PR #874, opened 2023-07-20
[Breaking Change] New builder pattern For Result

iNoles:
old way is 

```
loader().get(
    Request.Builder()
        .url(url)
        .parameters(parameters)
        .headers(headers)
        .build()
)
```

new way is 
```
loader().get {
    this.url = url
    this.parameters = parameters
    this.headers = headers
}
```

I think it is more simplified to cut down on verbose with more type-safe builder patterns. I am wondering if somebody may have a problem with this.


kittinunf (approved):
Lgtm!!

PR #881, opened 2023-11-14
Fixed YAML formatting

iNoles:
I used the yamllint with the same configuration that was set by Super linter to make sure it was completely accurate.


kittinunf (approved):
Looks great!

PR #905, opened 2024-05-19
[Fix] Fix publishing task to include sourceJa after verification process is failing

kittinunf:
### What's in this PR?

This PR fixes our gradle script a bit for publishing task when we publish the jar to sonatype.
Also it looks like we use `alpha0x` for naming convention so we are fixing that too.


iNoles (approved):
I hope it does add sourcejar correctly at this time with proper versions. Gradle seems to be okay with it too.

PR #907, opened 2024-05-19
Fix Gradle Actions Deprecation

iNoles:
This is coming from the https://github.com/gradle/actions/blob/main/docs/deprecation-upgrade-guide.md#the-action-gradlegradle-build-action-has-been-replaced-by-gradleactionssetup-gradle


kittinunf (approved):
👍

PR #908, opened 2024-05-20
Implement Response Headers

iNoles:
When I looked at Coil Alpha for the network modules, I noticed they use request and response for headers. Also, it would be useful for developers to inspect the HTTP Headers in their app too.


kittinunf:
Nice! Do you think we should add some tests to cover these?

iNoles:
> Nice! Do you think we should add some tests to cover these?

Sure, I can do that when I get a chance to do it.

kittinunf (approved):
Looks awesome! Let's do this

PR #921, opened 2024-06-22
I added 2 more tests for HttpLoaderFactoryTest and StringsTest

iNoles:
It is now

95% of Class
92.3% of Methods
89.2% of Line
90.8% of Instructions


kittinunf:
Nice!!

kittinunf (approved):

PR #925, opened 2024-06-28
Fixing some Github Actions files for the deprecated functionality 

iNoles:
Replace gradle/gradle-build-action with gradle/actions/setup-gradle by removing arguments so that we can use run for it.


kittinunf (approved):
Looks gtm!

PR #953, opened 2024-09-12
test release publish

jibae3:


iNoles:
Interesting, " If you want a workflow to run when stable and pre-releases publish, subscribe to published instead of released and prereleased." alpha is still pre-release.

PR #964, opened 2024-10-27
[New version] Release alpha04

kittinunf:
### What's in this PR?

Update and bump new version to alpha04


iNoles (approved):
I don't mind pushing this PR with/without markdown prettier. You can use https://prettier.io/ to fix it.

PR #981, opened 2025-01-07
Putting more libraries into version catalog

iNoles:
This PR is moving some libraries into the plugins section of version catalogs. It will make upgrading much easier.


kittinunf (approved):
+1 👍 great!

PR #993, opened 2025-03-12
Implement SSE Support

### LINKED ISSUE #989: Support `text/event-stream` requests
Duplicate of #723 

This library is really great, but there is no body streaming requests
I don't know if this is hard to implement since it's basically reading from the body stream until the client or the server stops the connection

Here's an example using java's `HttpClient`:
```kt
	HttpClient.newHttpClient().sendAsync(
		HttpRequest.newBuilder()
			.uri(URI.create("xxxxxx"))
			.header("Accept", "text/event-stream")
			.build(),
		HttpResponse.BodyHandlers.ofLines()
	).thenAccept { response ->
		response.body().forEach {
			if (!it.startsWith("data:")) return@forEach
                        println(it.substring(5).trim())
		}
	}.join()
```

iNoles:
Fix #989

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,2 @@
+[*/build/**/*]
+ktlint = disabled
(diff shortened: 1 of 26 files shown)


Edouard127:
Will this be backported to 2.x ?
3.x is not stable enough for us to use and still lacks proper documentation

iNoles:
> Will this be backported to 2.x ? 3.x is not stable enough for us to use and still lacks proper documentation

that is up to kittinunf for 2.x.

kittinunf (approved):
Looks great!

